PR #111, opened 2024-02-23
[WIP] Add PostgreSQL Support

saeedvaziry:


sineld:
The version has been configured as 14; however, it would be preferable if it were configurable.

saeedvaziry:
@sineld this commit is pretty old. the version will be configurable of course.

saeedvaziry:
Postgresql support added to 1.x via #118

Issue #140, opened 2024-03-27
Indicate Version on Dashboard


ebenwatts:
I will suggest you show the Vito version on the dashboard or footer of the project: Example: Vito v1.1.0

ebenwatts:
Seen, looks amazing

PR #146, opened 2024-03-29
Added footer & added version on config/app.php

saravanasai:
@saeedvaziry   I have created a footer & added version to config/app.php. As  #140 requested for this feature.

image


config/app.php
```diff
@@ -16,7 +16,15 @@
...
-
+     /*
+    |--------------------------------------------------------------------------
+    | Application Version
+    |--------------------------------------------------------------------------
+    |
+    | This value is the version of your application.
+    |
+    */
+    'version' => env('VITO_VERSION', 'v1.1.0'),
```
saeedvaziry, line 27: this means every time with every release this field needs to be updated. if you forget to update then again another release needs to be done.

saravanasai: @saeedvaziry   Yes, I thought to add a git pre-commit hook. But, I don't want to do any fancy items. Suggest to me if there is any better way to do it. Let me update this PR.

saeedvaziry: Thanks. Lets first discuss the solution. I would be in for something more automated

saravanasai: Sure. Let me check if is there any other solution & share you. we can discuss it.

saeedvaziry: Was also thinking of some solutions but looks a bit complicated.

PR #157, opened 2024-04-06
Feature/add application log viewer

borjajimnz:
This PR allow reading application logs using ssh cat on the server, specifying the full path and saving this path in sites.data_type. 

If something needs to be placed elsewhere, I can change it. I wasn't sure exactly where to place it. :)

Tests for the new route have been added.

Thank you very much for your work; I hope to continue collaborating on your impressive project.

Some screens:

Captura de pantalla 2024-04-07 a las 0 56 41

Captura de pantalla 2024-04-07 a las 0 57 27


saeedvaziry:
Thanks @borjajimnz for the contribution,
The coding and the way you added the feature completely fit with the rest of the app!
However, I have some concerns about the idea itself. What if some apps have multiple log files? like someone using Laravel's `daily` log driver.
Or what if the log file is huge enough that SSH will not be a suitable option to transfer them to the Vito and show it?

borjajimnz:
> Thanks @borjajimnz for the contribution,
> The coding and the way you added the feature completely fit with the rest of the app!
> However, I have some concerns about the idea itself. What if some apps have multiple log files? like someone using Laravel's `daily` log driver.
> Or what if the log file is huge enough that SSH will not be a suitable option to transfer them to the Vito and show it?

You right, let me find something better for this. Will come back with something better then. Thanks you. Regards

saeedvaziry:
@borjajimnz feel free to ping me on Discord. I also have some ideas for this.

PR #190, opened 2024-05-05
add nextjs site type

mehrdadx10:
add custom site type for **nextjs** projects

note:
only work for static exported projects
export folder is `out`

suggested deploy script
```
git pull
npm i
npm run build
npm run export
```


resources/views/application/nextjs-app.blade.php
```diff
@@ -0,0 +1,57 @@
+<div>
```
saeedvaziry, line 1: Can we do a refactor here? Since this file is kinda a copy of `php-app.blade.php`, we could create a `basic-app.blade.php` file and nextjs app and other apps using the `php-app.blade.php` can use it.
so this basic-app.blade.php will be a kinda layout that other apps can extend it!

saeedvaziry:
In general the PR looks good. The only problem I see here is that Vito doesn't install Node.js on the target servers!

To solve this problem I would recommend we make the Nodejs available to be installed as a service in the `Services > Supported Services` section.

resources/views/application/nextjs-app.blade.php
mehrdadx10: Sure, why not
In this pr, I wanted to do it with minimal changes.

mehrdadx10:
> In general the PR looks good. The only problem I see here is that Vito doesn't install Node.js on the target servers!
> 
> To solve this problem I would recommend we make the Nodejs available to be installed as a service in the `Services > Supported Services` section.



> In general the PR looks good. The only problem I see here is that Vito doesn't install Node.js on the target servers!
> 
> To solve this problem I would recommend we make the Nodejs available to be installed as a service in the `Services > Supported Services` section.

its seems  `app/SSH/OS/scripts/upgrade.sh` install nodejs.

app/ServerTypes/Nextjs.php
```diff
@@ -0,0 +1,80 @@
+<?php
```
saeedvaziry, line 1: This file supposed to be in the `App\SiteTypes` not `ServerTypes`

saeedvaziry:
OK. I figured it out! Supporting only the export mode in VitoDeploy doesn't really make sense.

I would go only if we're able to support nodejs server side applications.

However I don't want to close this PR. So lets change the direction to the `main` branch instead of `1.x` and work on it for the next major release.

Thanks for the work

saeedvaziry:
Closing due to inactivity

PR #216, opened 2024-05-16
Improve server console formatting

### LINKED ISSUE #209: Improve Console output format
Only 2 CSS rules for a much better formatting:

`font-family: monospace;
    white-space: pre;`

Before

image

After

image

alexanderzeillinger:
Closes #209

### DIFF
--- a/resources/views/console/index.blade.php
+++ b/resources/views/console/index.blade.php
@@ -65,7 +65,7 @@
 
         <div class="space-y-3">
             <x-console-view id="console-output">
-                <div class="w-full" x-text="output"></div>
+                <div class="w-full font-mono whitespace-pre" x-text="output"></div>
             </x-console-view>
             <form onsubmit="return false" id="console-form" class="flex items-center justify-between">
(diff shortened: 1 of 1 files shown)


resources/views/console/index.blade.php
```diff
@@ -65,7 +65,7 @@
 
         <div class="space-y-3">
             <x-console-view id="console-output">
```
saeedvaziry, line 67: could we have this in the component itself? `resources/views/components/console-view.blade.php`

resources/views/console/index.blade.php
```diff
@@ -65,7 +65,7 @@
...
-                <div class="w-full" x-text="output"></div>
+                <div class="w-full font-mono whitespace-pre" x-text="output"></div>
```
saeedvaziry, line 68: could we have this in the component itself? `resources/views/components/console-view.blade.php`

alexanderzeillinger:
At this point it's faster if you do it yourself.

saeedvaziry:
@alexanderzeillinger you were doing good tho. but sure I will handle it.

PR #218, opened 2024-05-21
Use Site PHP Version to Run Composer Install

devsarfo:
As mentioned by @maxcelos When creating a new site with the "Run composer install --no-dev" flag, it runs with the default PHP CLI, ignoring the PHP version selected in the form. It should use the selected PHP version.

This should pass the php_version to the getScript function to be used in the composer-install.sh

image
image

### DIFF
--- a/app/SSH/Composer/Composer.php
+++ b/app/SSH/Composer/Composer.php
@@ -14,6 +14,7 @@ class Composer
         $site->server->ssh()->exec(
             $this->getScript('composer-install.sh', [
                 'path' => $site->path,
+                'php_version' => $site->php_version,
             ]),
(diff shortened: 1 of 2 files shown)


saeedvaziry:
@devsarfo VitoDeploy doesn't set that kinda aliases for PHP binary.

image

devsarfo:
@saeedvaziry The change uses the php_version from the Site Model as shown below.
image

The version would be 8.2 instead of 82 and the command would run as shown below
image

saeedvaziry:
@devsarfo oh my bad. thats cool tnx

saeedvaziry (approved):

PR #242, opened 2024-06-20
refactor: :safety_vest: make name require & validation pattern to ipv…

youssefshibl:
set some of validation in ip input and port number and make name is require ,
this will help and to take off this work from backend and make check before send request to backend


saeedvaziry:
Backend validation is clean and easy to maintain. Also, it is centralized and we don't need to maintain both (front and back)

PR #243, opened 2024-06-22
Added Support for MariaDB Versions 10.6, 10.11, and 11.4

rtraselbd:
This update enhances the Ultimate Server Management Tool by adding support for MariaDB versions 10.6, 10.11, and 11.4.


saeedvaziry:
@rtraselbd thanks for the PR. will check it next week.

PR #249, opened 2024-07-01
Refactor validation rules to implement the new ValidationRule interface

bambamboole:
This Pr only does a bit of hygiene.
We remove the usages of the deprecated `Illuminate\Contracts\Validation\Rule` interface in favor of the new `Illuminate\Contracts\Validation\ValidationRule` interface.

Each rule is covered existing tests which can be verified by changing the validate methods temporally.


saeedvaziry:
Good catch!

PR #251, opened 2024-07-11
fixed issue using double quotes in .env

klc:
When double quotes are used in .env, when any update is save, the double quotes are deleted and the .env file is corrupted.

Before update
image

After update
image

### DIFF
--- a/app/Actions/Site/UpdateEnv.php
+++ b/app/Actions/Site/UpdateEnv.php
@@ -9,6 +9,7 @@ class UpdateEnv
     public function update(Site $site, array $input): void
     {
         $env = str_replace('"', '\"', $input['env']);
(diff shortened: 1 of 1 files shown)


saeedvaziry:
@klc I am not sure how this will behave in other scenarios but even if we need to do something like that I would recommend the `addslashes`  method rather than replacing double quotes!

Will try to reproduce the issue first.

klc:
@saeedvaziry  First, I tried the addslashes method, it solved the double quote problem, but when single quotes were used, the slash continued to remain after the update. Therefore addslashes is not the solution.

saeedvaziry:
Will fix (#259) the issue using uploads rather than playing with strings

PR #253, opened 2024-07-15
fix: vito agent localhost url issue

ahmetbedir:
### Issue: 

When .env APP_URL is empty, vito agent url is installed on the server as "http://localhost/api/servers/4/agent/33" as in the picture. Since it is installed in this way, it cannot send metrics. 

image

### Solution:

When installing vito agent service, config app.url is checked and if it is empty or localhost, temporary url is created with request schema and request url host and add this url into service type_data.

### DIFF
--- a/app/Actions/Service/Install.php
+++ b/app/Actions/Service/Install.php
@@ -28,7 +28,6 @@ class Install
         $service->type_data = $service->handler()->creationData($input);
         $service->save();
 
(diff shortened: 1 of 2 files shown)


saeedvaziry:
APP_URL should not be empty. Also the config file of the vito agent is editable so no need to fix it in the code.

PR #262, opened 2024-07-31
fix mixed content error

manusiakemos:
when i deploy on load balancer server it always give me error blocked mixed content. 
so i add config app.scheme to config/app.php to give user an option to force https url.
image


saeedvaziry:
I don't think this is needed. TrustProxy middleware should already solve the problem.
I will try to reproduce the issue first.

manusiakemos:
> I don't think this is needed. TrustProxy middleware should already solve the problem. I will try to reproduce the issue first.

sorry but in my case it's not enough with only TrustProxy middleware

manusiakemos:
the problem exist if i deploy laravel behind reverse proxy
https://github.com/livewire/livewire/discussions/3084#discussioncomment-8148210

saeedvaziry:
No plans for this at the moment

PR #267, opened 2024-08-03
Source Control Migration Fix

kkz6:
provider_data column in source_controls table is set as json. But in the model the cast is mentioned as encrypted array. So I have modified it to longtext in the new migration


saeedvaziry:
This is an unnecessary change and I would avoid it

PR #269, opened 2024-08-04
feat: Add Editor component for script, env or config editing

ahmetbedir:
### Deployment Script
image

### Env
image

### Nginx
image

### PHP.ini
image

### Scripts
image


saeedvaziry:
@ahmetbedir thanks for the contribution, If you check the history, Vito was using Ace editor in the beginning but because of some issues with interacting with Alpine and HTMX (in the past Livewire) I had to go back to the normal `textarea`
However, I will try to see if this works.

saeedvaziry:
Anyways, I think Monaco editor can give us more features like word based auto completion which might be super useful in the scripts.

@ahmetbedir Feel free to drop me a msg on discord so we can discuss.

ahmetbedir:
Thank you @saeedvaziry 

I didn't realise it had been used in the past. I'm now actively using Vito on my own server and thought a code editor would be very useful, so I wanted to send this PR.  

According to my tests, it has worked well so far. Maybe a few improvements may need to be made on it. 

I think Monaco Editor is more of a code editor than a script or env file editor. You can use it too, but my personal opinion Ace editor will be enough for small jobs. Laravel Forge uses Ace Editor in the same way.

ahmetbedir:
Also, if you do not change your mind about the Monaco Editor, I can update this PR for the Monaco Editor at my convenience. @saeedvaziry

PR #276, opened 2024-08-20
Update nginx.conf

kperrow:
added client_max_body_size 512M; 
Without this uploading anything on a WordPress site returns an NGINX error.


saeedvaziry:
1. the file you modified is the nginx config for the docker that we use to lunch vito locally.
2. I don't think it is a good practice to set 512MB limit on all wordpress sites.
3. If you want to change it you can modify the nginx config after you created the site using Vito and on the site settings page

PR #282, opened 2024-08-29
Feature/updated stalwart service and reverse proxy site

borjajimnz:
This PR will add the Stalwart-mail service, which is the best solution I’ve found so far for a complete email service: https://stalw.art/.

Additionally, when you install the Stalwart service this time, it will also automatically set up a reverse proxy site to access the Stalwart panel.

Since this type of site (reverse proxy) is now supported, I had to add a condition in the site configuration blade to only display the PHP version if it is not null.

Reverse proxy site type, could be also used to make Laravel Reverb apps, nodejs.. etc.

image

image

image

image

Stalwart panel:

image


borjajimnz:
Pint fails in github, but its working on local... different configurations?

image

saeedvaziry:
Thanks @borjajimnz, this will be a good addition.

Is the PR draft? I still see some unnecessary files like `.DS_Store` which you should not commit it.

Also I don't see any tests for this feature.

borjajimnz:
Yes, will open new PR, in the next days, thanks!!

saeedvaziry:
@borjajimnz why did you close the PR? :) I could help on this

PR #305, opened 2024-10-03
chore: delete `composer-setup.php` file after install completed in install script

datlechin:


saeedvaziry:
As already started on V2, we're not merging not-bug PRs to 1.x anymore.

here I tried to change the base branch but didn't work.

Feel free to raise another PR to the main branch.

Issue #321, opened 2024-10-21
[Vito v2.x] The installation script failed to complete database migrations on Ubuntu 20.04


joglomedia:
### Migration Failed on Ubuntu 20.04: `SQLSTATE[HY000]: General error: 1 Cannot add a NOT NULL column with default value NULL`

**Describe the bug**
Installing Vito v2.x on Ubuntu 20.04 using the standard installation script with the default SQLite3 library from the Ubuntu repository failed to complete the Laravel migrations.

In Connection.php line 813:
```SQLSTATE[HY000]: General error: 1 Cannot add a NOT NULL column with default value NULL (Connection: sqlite, SQL: alter table "storage_providers" add column "user_id" integer not null)```

In Connection.php line 565:
```SQLSTATE[HY000]: General error: 1 Cannot add a NOT NULL column with default value NULL```

**To Reproduce**
1. Create `Ubuntu 20.04` instance from your vps/cloud server provider (I use Amazon Lightsail)
2. Clone repo `git clone -b v2.x https://github.com/vitodeploy/vito.git`
3. Install Vito using default `scripts/install.sh`

**Expected behavior**
Laravel database successfully migrated, default admin user created.

**Screenshots**
![image](https://github.com/user-attachments/assets/4d771c25-6c9b-4867-a334-8430c43050e2)

**Additional context**
Operating System: 
```bash
$ sudo lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.6 LTS
Release:	20.04
Codename:	focal
```

SQLite3 library (default Ubuntu 20.04 repository packed with old version 3.31.1)
```bash
$ sudo apt-cache policy sqlite3
sqlite3:
  Installed: 3.31.1-4ubuntu0.6
  Candidate: 3.31.1-4ubuntu0.6
  ...
```
```bash
$ dpkg-query -l | awk '/sqlite3/'
ii  libsqlite3-0:amd64                 3.31.1-4ubuntu0.6                       amd64        SQLite 3 shared library
ii  php8.3-sqlite3                     8.3.12-1+ubuntu20.04.1+deb.sury.org+1   amd64        SQLite3 module for PHP
ii  sqlite3                            3.31.1-4ubuntu0.6                       amd64        Command line interface for SQLite 3
```

### Proposed Solution
The only solution I found was to build the latest version of SQLite3 (at least version 3.37.0, as used by Jammy) from source.
[...]

Issue #320, opened 2024-10-21
[Vito v2.x] The installation script is still referencing v1.x (latest release tag `1.11.0`)


joglomedia:
**Describe the bug**
Auto install v2.x installation failed. The installation script from `scripts/install.sh` is still referencing v1.x (latest release tag `1.11.0`)

**To Reproduce**
1. Create 'Ubuntu' instance from your vps/cloud server provider (I use Amazon Lightsail)
2. Execute `bash <(curl -Ls https://raw.githubusercontent.com/vitodeploy/vito/2.x/scripts/install.sh)`
3. See the installed VitoDeploy is version 1.x

**Expected behavior**
Running `scripts/install.sh` from v2.x should install Vito v2.x latest release

**Additional context**
The Git branching currently references a tag from the latest merged commit, but there's no released tag for v2.x.

`V_GIT_BRANCH=$(git tag -l --merged "${VITO_VERSION}" --sort=-v:refname | head -n 1)`

saeedvaziry:
branch 2.x doesn't have a release yet!

PR #326, opened 2024-10-24
Enhancement v2.x/scrips install.sh

joglomedia:
Hi all,
This PR contains significant changes, further review & comments are needed.

#### 1) Update `VITO_VERSION` from `1.x` to `2.x`
Issue #320

I added this validation to check the latest release tag and fall back to the version branch if it has not yet been released (it can be deleted after v2.x is officially released).

```bash
# Check for the latest version release tag
if [[ $(git tag -l --sort=-v:refname | head -n 1 | awk -F '.' '{print $1}') -eq $(echo "${VITO_VERSION}" | awk -F '.' '{print $1}') ]]; then 
  V_GIT_BRANCH=$(git tag -l --merged "${VITO_VERSION}" --sort=-v:refname | head -n 1)
else 
  V_GIT_BRANCH="${VITO_VERSION}" # If not available, fallback to the version branch
fi
```

#### 2) Vito install script requires root permissions
[Before] The script was executed by a non-root user, resulting in installation failure.
https://www.awesomescreenshot.com/video/32606272?key=ebaf580990e74287a71ed5af854fd96d

[After] Added root access validation and detect sudo user.
https://www.awesomescreenshot.com/video/32833984?key=22dfc7464ee02a559d412ee45c21ebb8

#### 3) Migration failure on Ubuntu 20.04
Issue #321

---

#### 1) Change Nginx repository from the OS default to `ondrej/nginx`
Since PHP already uses the `ondrej/php` repository, we can also use the `ondrej/nginx` repository to install the latest version of Nginx.

#### 3) Move directly configured sudo users from `/etc/sudoers` to `/etc/sudoers.d/90-vito-users`
Changes made directly to `/etc/sudoers` could be overwritten during system upgrades. In contrast, modifications made to files in `/etc/sudoers.d` remain intact when we upgrade the system. This can help prevent user lockouts during system upgrades.

#### 4) Update PHP from 8.2 to 8.3
Active support for PHP 8.2 will end at the end of this year. Meanwhile, Laravel 11 already supports PHP 8.3, so it is advisable to update.

#### 4) Add artisan commands `filament:optimize` and `filament:cache-components`
These commands enhance FilamentPHP's performance by caching components and optimizing them.

#### 5) Update Quick Start command to run install script

Install using the root user:
```sh
curl -sLo- https://raw.githubusercontent.com/vitodeploy/vito/2.x/scripts/install.sh | bash
```

Install with a sudo user:
```sh
curl -sLO https://raw.githubusercontent.com/vitodeploy/vito/2.x/scripts/install.sh && sudo bash ./install.sh
```

---

#### 1) Interactive (guided) and non-interactive install
Environment variables required for the non-interactive install:
```bash
NONINTERACTIVE=y
V_USERNAME=username
V_ADMIN_EMAIL=you@yourdomain.com
V_ADMIN_PASSWORD=yoursecurepassword
V_APP_ENV=production
V_USE_CUSTOM_DOMAIN=y/n
V_USE_HTTPS_DOMAIN=y/n
V_CUSTOM_DOMAIN=yourdomain.com
```

#### 2) Supported OS validation
At this time of writing, I suggest supporting the three latest major LTS releases of Ubuntu: Noble, Jammy, and Focal.

Video demo:
https://www.awesomescreenshot.com/video/32833293?key=ba8bf73e50564d51c5aee584b1ee4b33

#### 3) User input validation
https://www.awesomescreenshot.com/video/32834308?key=b13b0f720399d7e26138f5582ae4b471

#### 5) Enable Swap space for low-end box servers/VMs
It is required for SQLite3 compilation process on low RAM machine.

#### 7) Fancy header art
Inspired by DHH's [Omakub](https://github.com/basecamp/omakub/blob/master/ascii.sh) and my [LEMPer Stack](https://github.com/joglomedia/LEMPer/blob/master/scripts/utils.sh#L845) installer. ^^

image

### Interactive (Guided) Install
https://www.awesomescreenshot.com/video/32837448?key=fa8a1760bbf93922b48d2d6ff69ed66b

### Non-interactive (Unguided) Install
https://www.awesomescreenshot.com/video/32866352?key=a456e5e4fa0ab0e7cf29b7d429eb0d64


saeedvaziry:
Really appreciate the effort on this. But I liked it when it was simpler. 

It's a bit hard to review it now because of the many changes but I will take a look at it for sure.

saeedvaziry:
Still trying to understand some decisions here.

Vito is not going to support Redis for it's own cache driver. File driver works fine and not planning to make vito itself havy with requiring more resources to run it.

Why cloning the 2.x branch?! the branch will always get PRs merged into it but when it is stable we will release versions.

Tho I see some good features as well.

Maybe reach me on Discord to discuss this through.

joglomedia:
Hi @saeedvaziry, thanks for your time.

That's okay; I wasn't aware that Vito doesn't support Redis. We can cancel the Redis installation.

Cloning the 2.x branch is just a temporary fallback for the direct script installation test since we don’t have a released v2 tag yet. 

Once approved, I can revise the code.

Regards,

scripts/install.sh
```diff
@@ -1,215 +1,736 @@
...
+        grep -oE '[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}' | head -1)
+
+    echo "${SERVER_IP_PRIVATE}"
+}
+
+# Get server public IP Address
+function get_ip_public() {
+    local SERVER_IP_PRIVATE && SERVER_IP_PRIVATE=$(get_ip_private)
+    local SERVER_IP_PUBLIC && \
+    SERVER_IP_PUBLIC=$(curl -sk --ipv4 --connect-timeout 10 --retry 1 --retry-delay 0 https://freeipapi.com/api/json | tr -d '"' | awk -F "," '{print $2}' | awk -F "ipAddress:" '{print $2}')
```
saeedvaziry, line 213: You could just use `curl https://freeipapi.com` :)

scripts/install.sh
```diff
@@ -1,215 +1,736 @@
...
+    # Hack to detect public IP address behind NAT
+    if [[ "${SERVER_IP_PRIVATE}" == "${SERVER_IP_PUBLIC}" ]]; then
+        echo "${SERVER_IP_PRIVATE}"
+    else
+        echo "${SERVER_IP_PUBLIC}"
+    fi
+}
+
+# Get server private IPv6 Address
+function get_ipv6_private() {
```
saeedvaziry, line 224: I really don't think we need to fetch the ip addresses. I mean both ipv4 and ipv6. I still want to keep Vito ip agnostic

saeedvaziry: lets get rid of all IP calculations.

scripts/install.sh
```diff
@@ -1,215 +1,736 @@
...
+    if [[ $(groups "$(id -un)" | grep -c sudo) -ne 0 ]]; then
+      echo -e "\nFor a sudo user, you can run the following command:"
+      echo "curl -sLO https://raw.githubusercontent.com/vitodeploy/vito/${VITO_VERSION}/scripts/install.sh && sudo bash ./install.sh"
+    fi
+    exit 1
+  fi
+}
+
+# Check for supported OS
+function check_supported_os() {
```
saeedvaziry, line 170: This is a good addition! tnx

saeedvaziry:
@joglomedia I am still reviewing the PR, but definitely this is something we will have on v2. Tho I need some of them to be removed as well.

I don't want to support SSL generation out of the box. this will bring too many open issues as for any certbot failures! 

For SSL lets only rely on the documentation and the guides here (https://vitodeploy.com/introduction/securing.html)

So please remove all SSL related configs and just keep the certbot to be installed.

joglomedia:
> @joglomedia I am still reviewing the PR, but definitely this is something we will have on v2. Tho I need some of them to be removed as well.
> 
> I don't want to support SSL generation out of the box. this will bring too many open issues as for any certbot failures!
> 
> For SSL lets only rely on the documentation and the guides here (https://vitodeploy.com/introduction/securing.html)
> 
> So please remove all SSL related configs and just keep the certbot to be installed.

Hi @saeedvaziry thanks for the review, and I apologize for my late follow-up.

I have sent a new commit addressing the following:
- Removing SSL auto-configuration
- Removing IPv6 support
- Double-checking IPv4, as checking only the Ethernet interface for servers under NAT (such as AWS, Lightsail, and GCP) will only yield the private IP address
- Limiting the installer to support only Ubuntu LTS 24.04 (Noble) and 22.04 (Jammy)

saeedvaziry:
After a lot of back and forth, I decided to close this PR. 

I appreciate the time and effort you put on this however the PR brings a lot of changes and refactorings. The installer script currently is 200 lines and I would like to keep it simple for now.

PR #330, opened 2024-10-30
Add Browser Session Feature

rtraselbd:
This pull request introduces the Browser Session feature, which allows users to manage their browser sessions within their profiles. This enhancement aims to improve user experience by providing better session management capabilities.

Screenshot 2024-10-30 at 7 29 51 PM


saeedvaziry:
Thanks @rtraselbd for the effort. One of the biggest reasons that I decided to migrate to Filament was to avoid custom blade files like the one you have in this PR.

I believe such views still can be done via Filament, one example is the TwoFactor authentication feature in 2.x

rtraselbd:
Thanks, @saeedvaziry, I will try to implement it the Filament way. I appreciate your guidance!

saeedvaziry:
also we need tests for every feature. the test coverage should not go down

PR #339, opened 2024-11-04
Fix #279 - Privileges for `public` Schema in PostgreSQL 15+ 

### LINKED ISSUE #279: Postgres 16 - Database Owner Not Set
**Describe the bug**
When you create a new database in Postgres 16 and assign a user to it, it uses **postgres** as the owner instead of the one set.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a Server with Postgres as the DB
2. Go to databases 
3. Create a new database with a user
4. SSH into server 
5. Switch to  Postgres user using `sudo -i -u postgres`
6.  Access the PostgreSQL Command Line (psql) using `psql`
7. Run `\l` and observe the owner of the database you created

**Expected behaviour**
The Owner of the DB should be the one set in Vito and not posgres

**Screenshots**
Vito
image

psql
image

**Additional context**

devsarfo:
This PR checks the database version for PostgreSQL services and grants USAGE and CREATE permissions for the public schema if the version is 15 or higher.

This is required for Laravel migrations to work since, by default, the search_path for the `pgsql` driver is `public`

### DIFF
--- a/app/SSH/Services/Database/AbstractDatabase.php
+++ b/app/SSH/Services/Database/AbstractDatabase.php
@@ -117,6 +117,7 @@ abstract class AbstractDatabase extends AbstractService implements Database
     public function link(string $username, string $host, array $databases): void
     {
         $ssh = $this->service->server->ssh();
+        $version = $this->service->version;
 
         foreach ($databases as $database) {
             $ssh->exec(
@@ -124,6 +125,7 @@ abstract class AbstractDatabase extends AbstractService implements Database
                     'username' => $username,
                     'host' => $host,
                     'database' => $database,
+                    'version' => $version,
                 ]),
                 'link-user-to-database'
             );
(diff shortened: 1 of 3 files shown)


saeedvaziry:
Thanks @devsarfo , Please set the PR to 2.x as 1.x is not getting features and bug fixes anymore (only security patches can go to 1.x)

devsarfo:
Sure, I'll go ahead and do just that.

saeedvaziry:
@devsarfo are we closing this?

devsarfo:
I was hoping this fix would go into the 1.x also since it solves the issue which was before we branched out. I know a couple of users who are still using 1.x and this fix would come in handy.

saeedvaziry:
I see many requests coming for bug fixes on v1. since v2 is not fully ready, lets do one last patch release on v1

PR #345, opened 2024-11-04
fix: add needed folders

### LINKED ISSUE #343: Docker Start Script fails Due to Missing Folders
**Describe the bug**

I tried to run vitodeploy via docker swarm, then I've got "Please provide a valid cache path."

**To Reproduce**
1. Copy paste this docker-compose: https://vitodeploy.com/introduction/installation.html#docker-compose
2. Replace necessary environment, and additionally add APP_ENV=local APP_DEBUG=True
3. There should be no prior volumes, should start fresh.
4. After few seconds should show: "Please provide a valid cache path."

**Expected behavior**
Should run right away.

**Screenshots**
Screenshot 2024-11-04 at 19 20 42

**Additional context**
Manage to fixed by doing 

```sh
cd storage/
mkdir -p framework/{sessions,views,cache}
chown -R www-data:www-data framework
```

Maybe we could add it directly in start.sh

eznix86:
resolves #343

### DIFF
--- a/docker/start.sh
+++ b/docker/start.sh
@@ -54,8 +54,10 @@ if [ ! -f "$INIT_FLAG" ]; then
     touch "$INIT_FLAG"
 fi
 
-chown -R www-data:www-data /var/www/html \
-    && chmod -R 755 /var/www/html/storage /var/www/html/bootstrap/cache
+mkdir -p /var/www/html/storage/framework/{sessions,views,cache} &&
+    chown -R www-data:www-data /var/www/html &&
+    chmod -R 755 /var/www/html/storage /var/www/html/bootstrap/cache
+
 service php8.2-fpm start
 
 service nginx start


saeedvaziry:
But they are exist already

https://github.com/vitodeploy/vito/tree/1.x/storage/framework

eznix86:
@saeedvaziry that's correct, so i believe it should be moved to the script instead, because in your example you used named values which copies the files on first use, but on a file path, since the folder is empty, it will delete those files.

For example:

```yaml
services:
  vito:
    image: vitodeploy/vito:latest
    ports:
      - "8000:80"
    environment:
      #...
    volumes:
      - "./storage:/var/www/html/storage"
```


Maybe this below belongs to https://github.com/vitodeploy/vito/blob/1.x/docker/start.sh

```sh
mkdir -p /var/www/html/storage/framework/{sessions,views,cache} \
    && chown -R www-data:www-data /var/www/html \
    && chmod -R 755 /var/www/html/storage /var/www/html/bootstrap/cache
```

saeedvaziry:
@eznix86 no it doesn't stay on the container. in the docker-compose file if you do as the example says, it will create the volume itself.

jimblue:
@eznix86 I think what you looking for is:

```yaml
version: '3'
services:
  vito:
    image: vitodeploy/vito:latest
    environment:
      APP_KEY: ''
      NAME: ''
      EMAIL: ''
      PASSWORD: ''
    ports:
      - '6008:80'
    volumes:
      - 'vito-storage:/var/www/html/storage'
volumes:
  vito-storage:
    driver: local
    driver_opts:
      type: none
      device: /volume1/vitodeploy/storage
      o: bind
```

PR #357, opened 2024-11-10
Add Minio as storage provider

klc:
Hi
This PR introduces Minio as a new storage provider within the Vito Deploy system. Minio is an S3-compliant object storage service that you can install on your own server if you wish.


saeedvaziry:
@klc I appreciate the PR but we want to solve this problem for good by adding one S3-Compatible driver.

https://github.com/vitodeploy/vito/discussions/293

PR #408, opened 2024-12-24
phpstan 2.x

rudiedirkx:
Phpstan is very useful for finding flaws, even at level 2. Larastan makes it work for Laravel projects. Your IDE might be less happy, if it doesn't have great Laravel support like Larastan provides for Phpstan. I don't have a 'smart' IDE at all, so it works like a charm for me. Several dubious findings:

- Where do Server/ServerResource properties `auto_update` and `security_updates` come from? According to `Server` they're database columns, but I can't find them. Not even in old migrations/git history.
- The SSH helper doesn't know if it's an SSH or SFTP connection, because it's created before it's connected, and the `connect($sftp)` call decides which it's gonna be. A generic on the SSH class could help, but that feels odd. Feels like it should be 2 classes.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -42,3 +42,6 @@ jobs:
 
       - name: Run test suite
         run: php artisan test
+
+      - name: Run Phpstan
+        run: ./vendor/bin/phpstan analyze --no-progress
--- a/app/Actions/FirewallRule/CreateRule.php
+++ b/app/Actions/FirewallRule/CreateRule.php
@@ -21,7 +21,6 @@ class CreateRule
             'mask' => $input['mask'] ?? null,
         ]);
 
-        /** @var Firewall $firewallHandler */
         $firewallHandler = $server->firewall()->handler();
         $firewallHandler->addRule(
             $rule->type,
--- a/app/Actions/NotificationChannels/AddChannel.php
+++ b/app/Actions/NotificationChannels/AddChannel.php
@@ -59,7 +59,7 @@ class AddChannel
             'label' => 'required',
         ];
 
-        return array_merge($rules, static::providerRules($input));
+        return array_merge($rules, self::providerRules($input));
     }
 
     private static function providerRules(array $input): array
--- a/app/Actions/ServerProvider/CreateServerProvider.php
+++ b/app/Actions/ServerProvider/CreateServerProvider.php
@@ -17,7 +17,7 @@ class CreateServerProvider
      */
     public function create(User $user, Project $project, array $input): ServerProvider
     {
-        $provider = static::getProvider($input['provider']);
+        $provider = self::getProvider($input['provider']);
 
         try {
             $provider->connect($input);
@@ -60,7 +60,7 @@ class CreateServerProvider
             ],
         ];
 
-        return array_merge($rules, static::providerRules($input));
+        return array_merge($rules, self::providerRules($input));
     }
 
     private static function providerRules(array $input): array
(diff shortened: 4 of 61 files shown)


saeedvaziry:
Thanks for the effort. The docblocks in the Model classes are very useful for auto-completion. why did you remove them?

saeedvaziry:
Besides, I would start with phpstan itself instead of larastan and with level 1

saeedvaziry:
Just did a quick check on my local wth phpstan itself. found 9 errors on level 1 and 256 errors on level 2. shouldn't be difficult to fix them.

rudiedirkx:
> The docblocks in the Model classes are very useful for auto-completion. why did you remove them?

Because some of them were wrong (`array` instead of `Collection`, eg. https://github.com/vitodeploy/vito/blob/dfdd50beb7d95af2cf4ca7b75bc1e285cdbcc5e5/app/Models/Site.php#L295), and Larastan makes those better automatically. But if your IDE doesn't understand Laravel like Larastan does, this is a big miss, I get that.

> I would start with phpstan itself instead of larastan and with level 1

Phpstan without Larastan doesn't work, because Larastan is what makes sense of the loads of 'magic' in Laravel. There's also a level 0, but level 0 and 1 and 2 are all simple enough.

rudiedirkx:
> found 9 errors on level 1 and 256 errors on level 2. shouldn't be difficult to fix them.

Many of those 256 might be easy, and some of them will be impossible, because of Laravel's PHP magic. Larastan 'fixes' that with its Phpstan extensions. You also need fewer inline `@var` then.

saeedvaziry:
> Phpstan without Larastan doesn't work, because Larastan is what makes sense of the loads of 'magic' in Laravel. There's also a level 0, but level 0 and 1 and 2 are all simple enough.

I think its just configuration, it has worked on [Laravel](https://github.com/laravel/framework/blob/11.x/composer.json#L116) itself. should be easy to make it work on Vito as well.

rudiedirkx:
You can copy Larastan functionality, but why would you? That's why Larastan exists? It sounds like you're against Larastan specifically. Why? I like it. It covers all the Laravel magic, you can code like you're used to. You can even keep all the class `@property`, but they do have to be correct (`Collection` instead of `array` etc). Larastan won't overwrite manual `@property` tags. Downside of manual property tags is that they might be nonsense, and you won't know. For instance what's up with `auto_update` and `security_updates` in `Server`? Are those columns? Do they exist? They're in `$fillable` too, but not in any migration. Do they still exist, or is it just the `@property` tag that still exists?

Anyway, you're the boss, you decide if and how you want static analysis. I think it already found a few issues even at level 2.

rudiedirkx:
When the 2 `mixed` options are enabled, I expect many more errors, many of which aren't real errors, but Phpstan won't understand the code, because of Laravel magic handled by Larastan.

saeedvaziry:
Lets keep this open for a bit. Need to think more. Still have a bit of doubts of adding phpstan or not 🤔

rudiedirkx:
> Still have a bit of doubts of adding phpstan or not 🤔

It is a hassle getting it to work, but when it does it's nice to have an extra layer of protection.

saeedvaziry:
I totally agree with the need for such tool as the bigger the project goes and more contributed come in, we need more protection. I need to figure out what setup is the best for Vito.

saeedvaziry:
Alright, Back to this one. @rudiedirkx do you think you can boot up the PR again? we can go with Larastan 🚀

saeedvaziry:
Closing as stale

PR #413, opened 2024-12-30
UFW source mask 0 leads to 123.123.123.123/0 command that is not valid.

tobischulz:
In order to add new firewall rules without mask/subnet the interface requires a number for the mask field. Default is 0. Currently 0 leads to a ufw cli command like this:

```bash
sudo ufw allow from 123.123.123.123/0 to any proto tcp port 22
```

This command is not valid as the command returns:

```bash
vito@vito-012:~$ sudo ufw allow from 123.123.123.123/0 to any proto tcp port 22
WARN: Rule changed after normalization
Skipping adding existing rule
```

The code now checks if the mask is present and NOT 0 so it only append ~ /24 to the source IP when mask is set to a valid value.

With that change you are able to add new ufw rules on Ubuntu 24.04 and remove the default 22 from any rule.

### DIFF
--- a/app/SSH/Services/Firewall/Ufw.php
+++ b/app/SSH/Services/Firewall/Ufw.php
@@ -30,7 +30,7 @@ class Ufw extends AbstractFirewall
                 'protocol' => $protocol,
                 'port' => $port,
                 'source' => $source,
-                'mask' => $mask || $mask == 0 ? '/'.$mask : '',
+                'mask' => $mask || $mask != 0 ? '/'.$mask : '',
             ]),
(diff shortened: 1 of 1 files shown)


saeedvaziry:
Rule deletion doesn't work as expected

```
Could not delete non-existent rule

Firewall reloaded
```

saeedvaziry:
Also this change needs to work on the existing servers that has deployed 0.0.0.0/0

tobischulz:
I have created a new pull request with a better solution to this. Thanks

PR #415, opened 2024-12-30
Optional public ip config to add vito server to the firewall for created servers

tobischulz:
This will add a optional variable in the core config file to add the vito public ip automatically to the firewall for each created server.

Add the PUBLIC_IP to the .env file with the public IP of your vito installation. If set, after a new server is installed this IP will be added to the firewall as 22/tcp to be able to connect to all your servers even when the default rule 22/tcp any is removed manually.

This would be optional and does not change anything if not used. But it is helpfull for users that like to remove the open 22/tcp rule without the need to manually add this ip every time for each server.

By using this it could also help to not lock out of your own servers.

### DIFF
--- a/app/Actions/Server/CreateServer.php
+++ b/app/Actions/Server/CreateServer.php
@@ -218,12 +218,12 @@ class CreateServer
             ],
         ];
 
-        if (config('core.vito_public_up', false)) {
+        if (config('core.vito_public_ip', false)) {
             $defaultRules[] = [
                 'type' => 'allow',
                 'protocol' => 'tcp',
                 'port' => 22,
-                'source' => config('core.vito_public_up'),
+                'source' => config('core.vito_public_ip'),
                 'mask' => 0,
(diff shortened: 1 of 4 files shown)


app/Actions/Server/CreateServer.php
```diff
@@ -216,6 +216,19 @@ public function createFirewallRules(Server $server): void
                 'mask' => 0,
                 'status' => FirewallRuleStatus::READY,
             ],
-        ]);
+        ];
+
+        if (config('core.vito_public_up', false)) {
```
saeedvaziry, line 221: `up` looks like a typo no?

tobischulz: Sorry!! Will fix that :)

saeedvaziry:
Configs via env variables are not user friendly and I want to get rid of them.

This would be more practical feature after we implemented `Vito Settings` so admins can manage from dashboard.

For now lets just modify the rules after server creation.

PR #423, opened 2025-01-06
Feature/firewall default vito public ip

tobischulz:
This changes will make the firewall mask field nullable. On the add and remove rule the code checks for null or 0 and only append the subnet mask when $mask is not null.

With this the change existing servers are able to delete old rules and add new ones with or without masks.


tobischulz:
It was mixed with different prs that has been rejected. I will create a new pull request to fix the firewall rules without masks.

PR #431, opened 2025-01-12
Isolate Users

### LINKED ISSUE #429: Sites are not Isolated - sites have access to one another
**Describe the bug**
Deployed sites on a single server do not have the option to run on a different user account (all run on vito), therefore, one compromised site means that all sites are potentially compromised.   Sites should be able to be isolated, and therefore run under separate users accounts. 

**To Reproduce**
1. Create a PHP application site (site 1)
2. Create a file called "secrets.php", and store a secret in a variable
3. Create a PHP application site (site 2)
4. Upload a remote file editor to site 2
5. Browse to secrets.php from site 1 via the deployed script on site 2
6. See the secret values from site 1 from site 2

**Expected behavior**
Unable to access files from other sites

**Screenshots**
image
image

**Additional context**
N/A

RichardAnderson:
- [x] Isolate Site Username
- [x] Validation on Site Username
image
- [x] Creation of Isolated User on Isolation
- [x] Creation of FPM Pool on Isolation
- [x] Updated VHost to used isolated FPM pool when site is isolated
- [x] Split out the Path creation into a separate script
- [x] Deploy PhpMyAdmin under the isolated user account
- [x] Deploy Wordpress under the isolated user account
- [x] Deploy Laravel App under the isolated user account
- [x] Run Composer under the isolated user account
- [x] Run deployment script as isolated user
- [x] Allow configuration of queue based on isolated user
- [x] Allow isolated user accounts to be selected when running a script
image
- [x] Allow isolated user accounts to be used when setting up cron jobs
image
- [x] Run a script under an isolated user account
image
- [x] Update Broken Unit Tests
- [x] Create new Unit Tests (where required)

Resolves #429

### DIFF
--- a/app/Actions/Queue/GetQueueLogs.php
+++ b/app/Actions/Queue/GetQueueLogs.php
@@ -3,12 +3,11 @@
 namespace App\Actions\Queue;
 
 use App\Models\Queue;
-use App\Models\Site;
 
 class GetQueueLogs
 {
-    public function getLogs(Queue $queue, Site $site): string
+    public function getLogs(Queue $queue): string
     {
-        return $queue->server->processManager()->handler()->getLogs($site, $queue->getLogFile());
+        return $queue->server->processManager()->handler()->getLogs($queue->site, $queue->getLogFile());
     }
 }
--- a/app/Actions/Queue/GetQueueLogs.php
+++ b/app/Actions/Queue/GetQueueLogs.php
@@ -8,6 +8,6 @@ class GetQueueLogs
 {
     public function getLogs(Queue $queue): string
     {
-        return $queue->server->processManager()->handler()->getLogs($queue->site, $queue->getLogFile());
+        return $queue->server->processManager()->handler()->getLogs($queue->user, $queue->getLogFile());
     }
 }
--- a/app/Actions/Queue/GetQueueLogs.php
+++ b/app/Actions/Queue/GetQueueLogs.php
@@ -3,11 +3,12 @@
 namespace App\Actions\Queue;
 
 use App\Models\Queue;
+use App\Models\Site;
 
 class GetQueueLogs
 {
-    public function getLogs(Queue $queue): string
+    public function getLogs(Queue $queue, Site $site): string
     {
-        return $queue->server->processManager()->handler()->getLogs($queue->user, $queue->getLogFile());
+        return $queue->server->processManager()->handler()->getLogs($site, $queue->getLogFile());
     }
 }
--- a/app/Models/Queue.php
+++ b/app/Models/Queue.php
@@ -95,11 +95,7 @@ class Queue extends AbstractModel
 
     public function getLogDirectory(): string
     {
-        if ($this->user === 'root') {
-            return '/root/.logs/workers';
-        }
-
-        return '/home/'.$this->user.'/.logs/workers';
+        return '/home/'.$this->site->user.'/.logs/workers';
     }
 
     public function getLogFile(): string
--- a/app/Models/Queue.php
+++ b/app/Models/Queue.php
@@ -95,7 +95,11 @@ class Queue extends AbstractModel
 
     public function getLogDirectory(): string
     {
-        return '/home/'.$this->site->user.'/.logs/workers';
+        if ($this->user === 'root') {
+            return '/root/.logs/workers';
+        }
+
+        return '/home/'.$this->user.'/.logs/workers';
     }
 
     public function getLogFile(): string
--- a/app/SSH/Services/ProcessManager/Supervisor.php
+++ b/app/SSH/Services/ProcessManager/Supervisor.php
@@ -53,10 +53,11 @@ class Supervisor extends AbstractProcessManager
             ),
             true
         );
-        $this->service->server->ssh($user)->exec(
+        $this->service->server->ssh()->exec(
             $this->getScript('supervisor/create-worker.sh', [
                 'id' => $id,
                 'log_file' => $logFile,
+                'user' => $user,
             ]),
             'create-worker',
             $siteId
--- a/app/SSH/Services/ProcessManager/Supervisor.php
+++ b/app/SSH/Services/ProcessManager/Supervisor.php
@@ -2,6 +2,7 @@
 
 namespace App\SSH\Services\ProcessManager;
 
+use App\Models\Site;
 use App\SSH\HasScripts;
 use Throwable;
 
@@ -53,9 +54,10 @@ class Supervisor extends AbstractProcessManager
             ),
             true
         );
-        $this->service->server->ssh($user)->exec(
+        $this->service->server->ssh()->exec(
             $this->getScript('supervisor/create-worker.sh', [
                 'id' => $id,
+                'log_file' => $logFile,
             ]),
             'create-worker',
             $siteId
@@ -121,9 +123,12 @@ class Supervisor extends AbstractProcessManager
     /**
      * @throws Throwable
      */
-    public function getLogs(string $user, string $logPath): string
+    public function getLogs(Site $site, string $logPath): string
     {
-        return $this->service->server->ssh($user)->exec(
+        $ssh = $this->service->server->ssh();
+        if ($site->isIsolated()) { $ssh->asUser($site->user); }
+
+        return $ssh->exec(
             "tail -100 $logPath"
         );
     }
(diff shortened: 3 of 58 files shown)


RichardAnderson:
Further testing / work is required.

At the moment, sites (such as WordPress) are deployed as vito, they should be deployed as the isolated user (EG: You cant install plugins, as the isolated user that is running PHP does not have write access to the domain folder!).

Also need to run the deployment script as the isolated user as well.

database/migrations/2025_01_12_173141_add_isolated_user_to_sites.php
```diff
@@ -0,0 +1,30 @@
...
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('sites', function (Blueprint $table) {
+            $table->boolean('is_isolated')->default(false);
```
saeedvaziry, line 15: I wouln't differentiate the sites are isolated or not. I would just add one field `user` default to `vito`

and in every process I will take the user and do it based on that. so the system doesn't need to know if a site is isolated or not. it just needs to know what is the user of the given site.

app/Web/Pages/Servers/Sites/Pages/Queues/Index.php
```diff
@@ -47,17 +47,23 @@ protected function getHeaderActions(): array
                 ->label('New Queue')
                 ->form([
                     TextInput::make('command')
-                        ->rules(CreateQueue::rules($this->server)['command'])
+                        ->default(
+                            $this->site->type === 'laravel'
+                                ? "php{$this->site->php_version} {$this->site->path}/artisan queue:work"
+                                : ""
+                        )
```
RichardAnderson, line 54: Feedback here would be great.  I've added a default in when the type is laravel which defaults to the sites path with artisan queue:work.  I guess this should really be a seperate PR, let me know if you want me to split it out - happy to do so.

RichardAnderson: I've removed this, and will submit as another PR

saeedvaziry:
Wow 53 file changes is not something normally I would even go through :)

If you would avoid any extra changes that might not be related to the main goal of the PR (isolated users), what would you delete to make it smaller?

app/Actions/Queue/GetQueueLogs.php
```diff
@@ -3,11 +3,12 @@
 namespace App\Actions\Queue;
 
 use App\Models\Queue;
+use App\Models\Site;
 
 class GetQueueLogs
 {
-    public function getLogs(Queue $queue): string
+    public function getLogs(Queue $queue, Site $site): string
```
saeedvaziry, line 10: $queue already has the site

RichardAnderson:
I could likely remove the queue edits, cronjob edits and executions edits, but that would only result in a reduction of 5 files, and less than 40 lines of code.   Unfortunately, the remaining changes are vital to the goal of user isolation - it's quite a fundamental change.

app/Actions/Queue/GetQueueLogs.php
RichardAnderson: Removed this additional parameter and updated the calling code

saeedvaziry:
I'll be changing some stuff if you don't mind.

RichardAnderson:
@saeedvaziry No problem at all 👍

app/Models/Queue.php
```diff
@@ -95,11 +95,7 @@ public function site(): BelongsTo
 
     public function getLogDirectory(): string
     {
-        if ($this->user === 'root') {
-            return '/root/.logs/workers';
-        }
-
-        return '/home/'.$this->user.'/.logs/workers';
+        return '/home/'.$this->site->user.'/.logs/workers';
```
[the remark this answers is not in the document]
saeedvaziry: the root one is good addition, tho the user I think it should be the queue user not the site user. they might be same but still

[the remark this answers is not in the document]
saeedvaziry: the queue will be run as root user, and it will create log files in the home directory of the site's user. this means the site user has some files in its home that it doesn't have permissions to modify because the files belong to root.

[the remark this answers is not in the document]
RichardAnderson: As discussed, agree this could be a problem.  Therefore, suggest we define this as /home/{$this->user}/.logs/workers/{$this->id}.log.

That way, instead of creating a log file owned by root in the /home/example path, we will create it in the /home/root path.  If the queue is to be run as the example user, it will be created in the /home/example path, and vito /home/vito.

app/SSH/Services/ProcessManager/Supervisor.php
```diff
@@ -54,7 +53,7 @@ public function create(
             ),
             true
         );
-        $this->service->server->ssh()->exec(
+        $this->service->server->ssh($user)->exec(
             $this->getScript('supervisor/create-worker.sh', [
                 'id' => $id,
```
RichardAnderson, line 58: The isolated user won't have access to "sudo", so they won't be able to execute this script (create-workers.sh).  The "create-workers.sh" script needs to run as root.

app/SSH/Services/ProcessManager/scripts/supervisor/create-worker.sh
```diff
@@ -1,10 +1,6 @@
-if ! sudo mkdir -p "$(dirname __log_file__)"; then
-    echo 'VITO_SSH_ERROR' && exit 1
-fi
+mkdir -p "$(dirname __log_file__)"
 
-if ! sudo touch __log_file__; then
-    echo 'VITO_SSH_ERROR' && exit 1
-fi
+touch __log_file__
```
RichardAnderson, line 3: Given this script needs to be run as vito (so it can sudo to run supervisorctl) - we should continue to sudo these commands, as we cannot always assume that the vito has access to the isolated users directory

app/Support/helpers.php
```diff
@@ -178,3 +178,8 @@ function get_from_route(string $modelName, string $routeKey): mixed
 
     return null;
 }
+
+function escape_shell_arg(string $command): string
+{
+    return str_replace("'", "\\''", $command);
+}
```
RichardAnderson, line 185: Were we not replacing this with "addslashes"?

database/migrations/2025_01_12_173141_add_user_to_sites.php
```diff
@@ -13,13 +12,7 @@
...
     {
         Schema::table('sites', function (Blueprint $table) {
-            $table->string('user');
-        });
-
-        Site::with('server')->get()->each(function ($site) {
-            $site->update([
-                'user' => $site->server->getSshUser(),
-            ]);
+            $table->string('user')->default(config('core.ssh_user'));
```
RichardAnderson, line 15: This won't set "vito" on existing sites?

saeedvaziry: thats the reason. existing sites owned by vito

app/Support/helpers.php
saeedvaziry: forgot to revert this

app/SSH/Services/ProcessManager/Supervisor.php
saeedvaziry: Should be fine now

app/SSH/Services/ProcessManager/scripts/supervisor/create-worker.sh
saeedvaziry: Updated

database/migrations/2025_01_12_173141_add_user_to_sites.php
RichardAnderson: Agreed, however, this introduces inconsistencies - as newly created non-isolated sites will have a user of "vito" and old sites after being migrated will have a user of null.   Ideally, we'd want the user to be consistently vito, or consistently null

saeedvaziry: Default value is 'vito'

I think this will migrate old sites to have vito as user
New sites if they leave user as empty it will set vito

RichardAnderson: Fair enough - would be good to test this to be sure, I thought the default applied at SQL level, and therefore only took effect for new records.   It might depend on which DB engine you are using.

PR #433, opened 2025-01-15
Fix Deployment Script on Windows Clients

### LINKED ISSUE #428: Deploy Script Parsing New Lines
**Describe the bug**
A deployment script is treating new lines are part of the command.

**To Reproduce**
1. Use a Windows computer and Chrome
2. Create a Laravel Site and use the following deploy script:

> echo "Deploying $DOMAIN to $SITE_PATH"
> cd $SITE_PATH
> git fetch --all
> git reset --hard origin/$BRANCH
> git pull origin $BRANCH
> composer install --no-interaction --prefer-dist --optimize-autoloader --no-dev
> npm install
> php artisan migrate --force
> php artisan db:seed --force
> php artisan route:cache
> php artisan cache:clear
> php artisan view:clear
> npm run production
> echo "Site deployed!"

**Expected behavior**
Script runs each line as expected.

**Screenshots**
image
image
image

**Additional context**

RichardAnderson:
On windows browsers, when editing the deployment script \r\n was entered at the end of each line.  This caused an extra \r to be parsed by the linux server, causing deployment scripts to fail.

Resolves #428

### DIFF
--- a/app/Actions/Site/UpdateDeploymentScript.php
+++ b/app/Actions/Site/UpdateDeploymentScript.php
@@ -12,9 +12,9 @@ class UpdateDeploymentScript
      */
     public function update(Site $site, array $input): void
     {
-        $site->deploymentScript()->update([
-            'content' => $input['script'],
-        ]);
+        $script = $site->deploymentScript;
+        $script->content = $input['script'];
+        $script->save();
     }
 
     /**
--- a/app/Models/DeploymentScript.php
+++ b/app/Models/DeploymentScript.php
@@ -15,6 +15,15 @@ class DeploymentScript extends AbstractModel
 {
     use HasFactory;
 
+    protected static function boot(): void
+    {
+        parent::boot();
+
+        static::saving(function ($deploymentScript) {
+            $deploymentScript->content = str_replace("\r\n", "\n", $deploymentScript->content);
+        });
+    }
+
     protected $fillable = [
         'site_id',
         'name',
--- a/app/SSH/OS/OS.php
+++ b/app/SSH/OS/OS.php
@@ -7,6 +7,7 @@ use App\Models\Server;
 use App\Models\ServerLog;
 use App\SSH\HasScripts;
 use Illuminate\Filesystem\FilesystemAdapter;
+use Illuminate\Support\Facades\Log;
 use Illuminate\Support\Facades\Storage;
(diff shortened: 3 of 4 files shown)


app/SSH/OS/OS.php
```diff
@@ -162,7 +162,7 @@ public function runScript(string $path, string $script, ?ServerLog $serverLog, ?
         }
         $command = '';
         foreach ($variables as $key => $variable) {
-            $command .= "$key=$variable".PHP_EOL;
+            $command .= "$key=$variable\n";
```
RichardAnderson, line 165: \n should be used over PHP_EOL, as PHP_EOL can change with the running solution (On windows it will be \r\n, on Mac it could be \r or \n).  Linux, or more specifically Ubuntu utilises \n, so this should be hardcoded.

app/Actions/Site/UpdateDeploymentScript.php
```diff
@@ -12,9 +12,9 @@ class UpdateDeploymentScript
      */
     public function update(Site $site, array $input): void
     {
-        $site->deploymentScript()->update([
```
saeedvaziry, line 15: any specific reason for this change?

app/Models/DeploymentScript.php
```diff
@@ -15,6 +15,15 @@ class DeploymentScript extends AbstractModel
 {
     use HasFactory;
 
+    protected static function boot(): void
+    {
+        parent::boot();
+
+        static::saving(function ($deploymentScript) {
+            $deploymentScript->content = str_replace("\r\n", "\n", $deploymentScript->content);
```
saeedvaziry, line 23: why not just simply doing inside the action?

RichardAnderson: Provides the functionality on the model directly, so anywhere else we want to set the value or edit it (for example, an API, or quick action, etc), the functionality always fires.  Avoids duplications in the future.

saeedvaziry:
Although Vito is not built for Windows but if this tiny change will make it work there why not

RichardAnderson:
@saeedvaziry - Just to note, this issue was when setting the deploy script via a browser running windows, not running vito on windows.    Setting the script on my windows machine via Chrome results in a failure, setting the script on my Mac running Chrome results in a success.

PR #434, opened 2025-01-15
Define a default queue command for laravel projects

RichardAnderson:
For Laravel projects, define a default command for queues.


saeedvaziry:
I know this might be handy but Vito is not built to be opinionated and I won't add such hardcoding feature for the sake of keeping the codebase smaller and maintainable

marktopper:
It could be a config?

saeedvaziry:
We have some env variables available for deployment scripts.

https://vitodeploy.com/sites/application#deployment-script

I was thinking maybe we can have such variables and most of the commands and scripts we have.

PR #435, opened 2025-01-15
VHost Templating using Twig

### LINKED ISSUE #303: Force HTTPS configuration for sites
### Discussed in https://github.com/vitodeploy/vito/discussions/275

<sup>Originally posted by **agustinzamar** August 14, 2024</sup>
I already deploployed a site and created a ssl certificate for it. How can I force redirect to https site when the user access to http? I tried manually updating the vhost but if there are any changes it will get reset I believe</div>

RichardAnderson:
## Summary
Allows dynamic VHost configuration using Twig template engine.  

This means a single VHost template can be used, and it supports all languages, HTTP/HTTPS and forced HTTPS without wiping out the users changes each deploy/change. 

image

### Site Creation
image

### Update VHost
image

## Task List
- [x] Deploys a VHost from the twig template
- [x] Stores the sites template
- [x] Builds the template on update
- [x] Validates the build VHost file before deploying it
- [ ] Rebuild VHost on SSL Enable / Disable
- [ ] Migration to store "force https" setting on the site
- [ ] UI to allow setting of force of HTTPS using redirect in the template
- [ ] Remove "force https" when SSL removed
- [ ] Update UI to define it as a template

Resolves #303 

# Design Decisions
## Twig vs Blade

When users need to create or edit templates in PHP/Laravel apps, Twig is generally the safer choice than Blade because it runs in a protected environment and lets you control exactly what users can access. Unlike Blade, which allows direct PHP execution and unrestricted access to functions and variables, Twig keeps things locked down by default.

1. **Sandboxing**
Twig is designed with sandboxing in mind and can strictly control what users can do in templates
Blade is designed to be a full-featured template engine with access to PHP, making it riskier for user templates

2. **PHP Access**
Twig by default doesn't allow any PHP execution
Blade allows direct PHP execution with php directives, making it dangerous for user templates, it's not trivial to disable this

3. **Function Access**
Twig lets you explicitly whitelist which functions users can access
Blade has access to all PHP functions by default, which is a major security risk for user provided/edited templates

4. **Variable Handling**
Twig implements a strict system where variables must be explicitly passed to templates and undefined variables trigger errors by default. In contrast, Blade offers a more flexible approach, allowing direct access to any variables within the current scope, including global variables, and takes a more permissive stance when dealing with undefined variables.


saeedvaziry:
I will come back to this later but why would we twig while we have blade?

RichardAnderson:
Very good question - the answer is security.  Blade is great for static templates, but for dynamic templates, we don't want the "power" of blade and it's php + other directive.   Twig is lightweight and secure, and works without needing to "hack" out any unsecure functions.

RichardAnderson:
Will pause on this one until we've had the opportunity to discuss further.  Feel free to message/call me on Discord when you are available.

PR #436, opened 2025-01-16
Edit & Download (local) Backups

### LINKED ISSUE #425: Backups download and edit
### Discussed in https://github.com/vitodeploy/vito/discussions/424

<sup>Originally posted by **miquelangeld** January  7, 2025</sup>
It would be quite useful to be able to download the backups from the database page at least for local backups.
Also it would be great the possibility to modify the interval of a backup.

image
</div>

RichardAnderson:
## Summary
Provides the ability to edit backup configuration through a new action on the ListBackups table.  Once clicked, this opens a form to allow the editing of the Interval and number of backups to keep. 
image
image
image

Provides the ability to download a backup (assuming it's stored locally) via the ListBackupFiles table.
image
image
image

Resolves #425 

## Tasks
- [x] Add edit backup icon to Backup list
- [x] Add dialog form to edit interval and backups to keep
- [x] Update backup in the database
- [x] Create/update unit tests for updating a backup
- [x] Add a download backup icon to the Backup file list
- [x] Ensure download backup icon is only visible for local files
- [x] Start a download of the backup when the download backup action is clicked
- [x] Purge locally downloaded backups after 24 hours
- [x] On delete of a backup file, delete the backup file on the server
- [x] Upgrade S3 handler / service from v1 to v2
- [x] Ability to delete an S3 file
- [x] On delete of the backup configuration, delete all backup files on the given backup
- [x] Update description for delete of backup configuration to define it will delete the files

# Design Decisions
## Downloading Files
After trying a few different approaches to download the backup, I settled on downloading the backup from the remote server via SFTP, and then streaming the output to the user.  This provides better support than just downloading into memory and sending to the client, as we can get out of memory exceptions or hit PHP memory limits when dealing with larger files.  

This implementation seems to work on moderately sized databases - I tried on a 1GB database (zipped down to 200mb), and it transfers pretty fast to the server and streams to the client.  There is obviously a timeout on the webserver, so if the file is too large, or the connection from the Vito server to remote server is poor, the request might time out.

The only way to resolve the above would be to offload the download of the zip file to queue, however, this would require a lot more work in vito, and if the file is that large, then users should be downloading it through other means than a web interface! 

## S3 Upgrade to v2
In order to support more providers, we've upgraded the S3 AWS CLI to version 2, however, the latest version of the AWS CLI has an issue with Hetzner, DigitalOcean and a few other providers whom have yet to implement checksum validation on their implementations of S3.  

If you are getting an error that signing requests have failed, such as `XAmzContentSHA256Mismatch` then you will need to downgrade the AWS CLI to 2.22.35 until your S3 backup provider has implemented checksum validation.  You can do this via a script run on the server as root:

```
rm -rf /usr/local/bin/aws /usr/local/aws-cli/
ARCH=$(uname -m)
curl "https://awscli.amazonaws.com/awscli-exe-linux-$ARCH.zip" -o "aws.zip"
unzip -q aws.zip
./aws/install --bin-dir /usr/local/bin --install-dir /usr/local/aws-cli
rm -rf aws aws.zip
```

In order to remove the outdated version of AWS CLI (2.22.35), you can run the following script as root:

```
rm -rf /usr/local/bin/aws /usr/local/aws-cli/
```

## S3 / Dropbox deletion
Historically, API calls were made via the vito server to delete dropbox files for the example, however, to support WAFs and IP restrictions, now, the commands to delete the backups are run on the server.  This means access to FTP, Dropbox, S3, etc. can be restricted to only the servers that need access should that be required. 

At present, the initial setup/connection is through the vito server, I'd look to add a new PR in the future to replace with a "Test" button, and allow the user to submit the details without verification, or test the connection first.


RichardAnderson:
Not happy with the download functionality... Will refactor slightly to make it more resilient.

RichardAnderson:
Download functionality is not perfect, but it works for smaller files. 
I'll look to make it more resilient in future PRs

saeedvaziry:
facing a weird issue when uploading to AWS S3. not sure is about this PR or not
```
Completed 616 Bytes/616 Bytes (3.2 KiB/s) with 1 file(s) remaining
upload failed: ./app-20250125154721.zip to s3://vito-test/app/app-20250125154721.zip An error occurred (PermanentRedirect) when calling the PutObject operation: The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint: vito-test.s3-ap-southeast-1.amazonaws.com
You can fix this issue by explicitly providing the correct region location using the --region argument, the AWS_DEFAULT_REGION environment variable, or the region variable in the AWS CLI configuration file. You can get the bucket's location by running "aws s3api get-bucket-location --bucket BUCKET".
```

saeedvaziry:
was an issue with the bucket name. no problem

PR #437, opened 2025-01-17
Filter out project users

### LINKED ISSUE #421: Do not suggest adding a user to the project who is already in the project
When I'm on the Project editing screen, there is a "Add user" section.
When I click inside of it - I see the current user (myself) as well, who is already in this project (which is clearly visible below in the table of project users.

Selecting that user results in an error: 

image

Users that are already in the project should not appear in this dropdown:
image

RichardAnderson:
Ensures that only users not already within a project are suggested for the add user dropdown.

Resolves #421 

@saeedvaziry a small PR for you... finally!


saeedvaziry:
I love small PRs :)

PR #458, opened 2025-02-02
Built-in File Manager

saeedvaziry:
This version is literally over SSH. need to check the SFTP feature of phpseclib

### DIFF
--- a/app/Web/Pages/Servers/FileManager/Widgets/FilesList.php
+++ b/app/Web/Pages/Servers/FileManager/Widgets/FilesList.php
@@ -269,10 +269,9 @@ class FilesList extends Widget
             ->after(function (array $data) {
                 run_action($this, function () use ($data) {
                     foreach ($data['file'] as $file) {
-                        $this->server->os()->write(
+                        $this->server->ssh($this->serverUser)->upload(
+                            Storage::disk('tmp')->path($file),
                             $this->path.'/'.$file,
-                            Storage::disk('tmp')->get($file),
-                            $this->serverUser
                         );
                     }
                     $this->refresh();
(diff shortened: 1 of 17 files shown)


app/Web/Pages/Servers/FileManager/Widgets/FilesList.php
```diff
@@ -0,0 +1,371 @@
...
+    protected function downloadAction(): Action
+    {
+        return Action::make('download')
+            ->tooltip('Download')
+            ->icon('heroicon-o-arrow-down-tray')
+            ->hiddenLabel()
+            ->visible(fn (File $file) => $file->type === 'file')
+            ->action(function (File $file) {
+                $file->server->ssh($file->server_user)->download(
+                    Storage::disk('tmp')->path($file->name),
```
RichardAnderson, line 315: tmp will be shared across multiple users / sites running similar processes.  Filenames should be relatively unique, it would not be advisable to store simple named files in tmp.

RichardAnderson:
General Feedback:

We are currently using a database table to extract and load the files into a filament table.   However, this is pretty wasteful and not performant.  Ideally, we should use a package such as "Sushi" (The in-memory array connection for Laravel) to keep the data in-memory, and cache the data for future requests. 

We'd also want to define a "session" when the filament control is mounted - one that can be used to pull back this sessions files, rather than all files of the user, this would allow a user to have multiple active tabs or browser sessions of the file manager without causing problems. 

Loading the content into an in-memory adapter like Sushi, and using a cache to store this information against a session identifier would increase performance and lower database storage (as we can expire the cache after an hour or two automatically), while keeping with filaments table functionality.

PR #461, opened 2025-02-04
Fix missing ip for AWS and DO

### LINKED ISSUE #459: Deploying servers through DigitalOcean fails
**Describe the bug**
When trying to create a new server using DigitalOcean as provider, it seems to always fail (status installation_failed). The issues seems to be that VitoDeploy does not find the IP address of the created DigitalOcean server. The droplet exists in the DigitalOcean panel, and if I manually copy the IP to VitoDeploy, it finds the server and changes the status to ready (although there is a mismatch between installed services/expected services).

**To Reproduce**
1. Go to Servers,
2. Click on Create a Server,
3. Create a server with DigitalOcean as a provider,
4. Wait for the installation to complete,
5. The installation never gets further than 0%, eventually ending up with "installation_failed" status.

**Expected behavior**
The server should be created and provisioned.

saeedvaziry:
closes #459

### DIFF
--- a/app/Actions/Server/CreateServer.php
+++ b/app/Actions/Server/CreateServer.php
@@ -86,6 +86,9 @@ class CreateServer
             while ($maxWait > 0) {
                 sleep(10);
                 $maxWait -= 10;
+                if (! $server->provider()->isRunning()) {
+                    continue;
+                }
                 try {
                     $server->ssh()->connect();
                     break;
--- a/app/ServerProviders/AWS.php
+++ b/app/ServerProviders/AWS.php
@@ -149,6 +149,10 @@ class AWS extends AbstractProvider
                 $this->server->save();
             }
 
+            if (! $this->server->ip) {
+                return false;
+            }
+
             if (isset($result['Reservations'][0]['Instances'][0]['State']) && isset($result['Reservations'][0]['Instances'][0]['State']['Name'])) {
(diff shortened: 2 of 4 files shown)


austinkregel:
👀 I just tried this fix and wanted to report it does fix the issue creating Digital Ocean servers;

However, I also wanted to report this issue impacts Vultr as well 😞

saeedvaziry:
@austinkregel have you by any chance configured vito's supervisor to have more processes? by default it has 1

saeedvaziry:
The installed has been configured to install everyone in only one queued job because if 2 processes use `apt` to install something it can end up with the error in the screenshot. which I don't think is the case of this issue

austinkregel:
> @austinkregel have you by any chance configured vito's supervisor to have more processes? by default it has

I hadn't configured anything extra, Vito Deploy had only been installed on a fresh machine a few hours earlier. I added my DO credentials and selected a small droplet.

I suspect my droplet was too small and things took too long.

RichardAnderson:
> > 👀 I just tried this fix and wanted to report it does fix the issue creating Digital Ocean servers; though it did fail to get a lock while trying to install supervisor and failed, though deleting the server and trying again succeeded image
> 
> I have some issue in this version

What size of droplet did you use?   At minimum you'd need the 1GB RAM, however 2 GB would be recommended.

Issue #469, opened 2025-02-07
Unable to create mariadb user after the deployment of a new server


austinkregel:
**Describe the bug**

**To Reproduce**
1. Deploy a new server with MariaDB 11
2. Deploy a new site with a new database & user.
3. Fill out the fields, and create.
4. Page should error
5. In the logs you should see something like:

```
[2025-02-07 03:27:59] production.ERROR: View [ssh.services.database.mariadb.create] not found. {"userId":1,"exception":"[object] (InvalidArgumentException(code: 0): View [ssh.services.database.mariadb.create] not found. at /home/vito/vito/vendor/laravel/framework/src/Illuminate/View/FileViewFinder.php:139)
[stacktrace]
#0 /home/vito/vito/vendor/laravel/framework/src/Illuminate/View/FileViewFinder.php(79): Illuminate\\View\\FileViewFinder->findInPaths()
#1 /home/vito/vito/vendor/laravel/framework/src/Illuminate/View/Factory.php(151): Illuminate\\View\\FileViewFinder->find()
#2 /home/vito/vito/vendor/laravel/framework/src/Illuminate/Foundation/helpers.php(1062): Illuminate\\View\\Factory->make()
#3 /home/vito/vito/app/SSH/Services/Database/AbstractDatabase.php(89): view()
#4 /home/vito/vito/app/Actions/Database/CreateDatabase.php(21): App\\SSH\\Services\\Database\\AbstractDatabase->create()
#5 /home/vito/vito/app/Web/Pages/Servers/Databases/Index.php(64): App\\Actions\\Database\\CreateDatabase->create()
#6 /home/vito/vito/app/Support/helpers.php(68): App\\Web\\Pages\\Servers\\Databases\\Index->App\\Web\\Pages\\Servers\\Databases\\{closure}()
#7 /home/vito/vito/app/Web/Pages/Servers/Databases/Index.php(63): run_action()
#8 /home/vito/vito/vendor/filament/support/src/Concerns/EvaluatesClosures.php(35): App\\Web\\Pages\\Servers\\Databases\\Index->App\\Web\\Pages\\Servers\\Databases\\{closure}()
#9 /home/vito/vito/vendor/filament/actions/src/MountableAction.php(41): Filament\\Support\\Components\\Component->evaluate()
#10 /home/vito/vito/vendor/filament/actions/src/Concerns/InteractsWithActions.php(98): Filament\\Actions\\MountableAction->call()
[...]

PR #475, opened 2025-02-07
Add missing views for Mariadb

austinkregel:
This fixed https://github.com/vitodeploy/vito/issues/469


austinkregel:
I didn't realize renaming the branch would close an open PR. I'll create another PR 😅

PR #488, opened 2025-02-15
Firewall Refactor

RichardAnderson:
Updates the firewall functionality to support an easier to use creation interface, and more resilient application process.  Includes the ability to edit existing rules.

Updated List View:
image

Updated Creation Interface:
Includes quick action to insert current users IP address into the source
image
image

New Edit Interface:
image


database/migrations/2025_02_15_082027_update_firewall_rules_table.php
```diff
@@ -0,0 +1,39 @@
...
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('firewall_rules', function (Blueprint $table) {
+            $table->string('name')->default('Undefined')->after('id');
+            $table->ipAddress('source')->default(null)->nullable()->change();
+        });
+
+        DB::statement("UPDATE firewall_rules SET name = UPPER(protocol) WHERE protocol IN ('ssh', 'http', 'https')");
```
saeedvaziry, line 19: what if I had other rules?

database/migrations/2025_02_15_082027_update_firewall_rules_table.php
```diff
@@ -0,0 +1,39 @@
...
+        DB::statement("UPDATE firewall_rules SET protocol = 'tcp' WHERE protocol IN ('ssh', 'http', 'https')");
```
saeedvaziry, line 20: aren't these already `tcp`?

database/migrations/2025_02_15_082027_update_firewall_rules_table.php
```diff
@@ -0,0 +1,39 @@
...
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('firewall_rules', function (Blueprint $table) {
+            $table->string('name')->default('Undefined')->after('id');
```
saeedvaziry, line 15: not a blocker but `->after...` not gonna work in SQLite

app/Enums/FirewallRuleStatus.php
```diff
@@ -2,11 +2,31 @@
 
 namespace App\Enums;
 
-final class FirewallRuleStatus
+enum FirewallRuleStatus: string
```
saeedvaziry, line 5: Lets please not change this to enums yet. We can do a proper refactor on this later. I don't want to have enums partially as enum and partially as classes.

RichardAnderson: Do I read that as an indication that you are happy that they all change to this format? (I did not change them all to reduce the PR size).  I can 100% submit a PR to change them all.

database/migrations/2025_02_15_082027_update_firewall_rules_table.php
RichardAnderson: Unfortunately not - they are created with the protocol as 'ssh', 'http' and 'https', and "getRealProtocol" defaults to tcp if its not udp.

database/migrations/2025_02_15_082027_update_firewall_rules_table.php
RichardAnderson: Other rules the protocol is limited to tcp or udp, so not a problem, it's ONLY an issue on the system created rules, which sets the protocol to something other than 'tcp' or 'udp'

resources/views/ssh/services/firewall/ufw/apply-rules.blade.php
```diff
@@ -0,0 +1,37 @@
+@include('ssh.services.firewall.ufw.backup-rules')
+
+if ! sudo ufw --force reset; then
+    echo 'VITO_SSH_ERROR' && exit 1
+fi
+
+
+if ! sudo ufw default deny incoming; then
```
saeedvaziry, line 8: This makes me really scared. Imagine if something happens after this command. Are we going to lose access to the server?

RichardAnderson: If someone removes port 22 - potentially (but they already would!), but if the rules fail to apply, it automatically restores the backup.  This is why the rollback was put in place.

PR #489, opened 2025-02-15
Database collations

RichardAnderson:
Pulls the collation and charset information from the database on install.
Syncs the charset information to existing DBs
Provides the ability to specify the charset.
Partially implements synching of DBs.

image
image

- [x] Test with MySQL
- [x] Test with MariaDB
- [x] Test with PostgreSQL

image


database/migrations/2025_02_15_120213_update_databases_table.php
```diff
@@ -0,0 +1,48 @@
...
+            $service = $server->defaultService('database');
+
+            if (! $service) {
+                continue;
+            }
+
+            /** @var Database $db */
+            $db = $service->handler();
+
+            $db->syncDatabases(false);
```
saeedvaziry, line 33: We need a rollback mechanism and let the user to retry if this fails

app/Models/Database.php
```diff
@@ -48,12 +52,24 @@ public static function boot(): void
...
-    public static array $statusColors = [
-        DatabaseStatus::READY => 'success',
-        DatabaseStatus::CREATING => 'warning',
-        DatabaseStatus::DELETING => 'warning',
-        DatabaseStatus::FAILED => 'danger',
-    ];
+    public function getStatusText(): string
+    {
+        return match ($this->status) {
+            DatabaseStatus::READY => 'Ready',
```
saeedvaziry, line 58: Please don't 😂

saeedvaziry:
@RichardAnderson I added some unit tests. they might fail now but it is expected to pass

saeedvaziry:
@slaFFik this PR uses `utf8mb3` as default choice already.

RichardAnderson:
> `utf8mb3` is being phased out in MySQL, so it's better to switch to `utf8mb4` for future compatibility.
> 
> Also, iirc emojis can't be stored correctly with `utf8mb3`, plus there are issues with some Chinese/Japanese/Korean characters.

This is true, however, we are attempting to support 3 versions of MySQL, so uft8mb4 is not supported on all of them.   I guess for later version we should change the default collation, but at the moment, nothing in the service can identify the version, it's treated generically.

PR #496, opened 2025-02-16
Fix .env Files for Isolated Users

### LINKED ISSUE #495: Cannot update .env
**Describe the bug**
.env is not being updated when the site is isolated.

**To Reproduce**
1. Create an isolated site with Laravel type for example
2. Try to update .env file

RichardAnderson:
Resolves #495

### DIFF
--- a/app/Helpers/SSH.php
+++ b/app/Helpers/SSH.php
@@ -112,7 +112,9 @@ class SSH
 
         try {
             if ($this->asUser) {
-                $command = 'sudo su - '.$this->asUser.' -c '.'"'.addslashes($command).'"';
+                $command = addslashes($command);
+                $command = str_replace('\\\'', '\'', $command);
+                $command = 'sudo su - '.$this->asUser.' -c '.'"'.trim($command).'"';
             }
 
             $this->connection->setTimeout(0);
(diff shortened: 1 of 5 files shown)


app/Helpers/SSH.php
```diff
@@ -112,7 +112,9 @@ public function exec(string $command, string $log = '', ?int $siteId = null, ?bo
 
         try {
             if ($this->asUser) {
-                $command = 'sudo su - '.$this->asUser.' -c '.'"'.addslashes($command).'"';
+                $command = addslashes($command);
+                $command = str_replace('\\\'', '\'', $command);
+                $command = 'sudo su - '.$this->asUser.' -c '.'"'.trim($command).'"';
```
RichardAnderson, line 117: Running the command within double quotes, means we don't need to escape single quotes, as these will not be considered as an escape from the double quotes.  This allows us to parse files effectively through SSH instead of SFTP, saving us a connection and a headache with access.

RichardAnderson:
> @RichardAnderson were you able to reproduce the issue?

I'm able to replicate on my current system and a brand new system.  I added a log statement for SFTP errors, and I get the following error:

[2025-02-21 17:00:57] local.INFO: sftp errors ["NET_SFTP_STATUS_PERMISSION_DENIED: Permission denied"] 

Video here:  https://share.cleanshot.com/mszZTdxK

PR #510, opened 2025-02-28
fix: add missing php-intl package when installing php

dyanakiev:
Added `php{{ $version }}-intl` to the php installation list as it was missing. This package is a standard component in most projects and should be included by default.


saeedvaziry:
Although you can install it later via extensions using Vito but this is ok.

dyanakiev:
> Although you can install it later via extensions using Vito but this is ok.

Yes, I've noticed that when creating a new site from a repository and running `composer install`, it often fails due to this issue and there is no way to fix the status of the site or retry it. For instance, if a Laravel project includes Filamentphp which required php intl.

PR #513, opened 2025-02-28
feat: add editable site type field with modal action

dyanakiev:
This improves flexibility by allowing users to correct or update the site type (e.g., Laravel, WordPress), ensuring access to the appropriate features which Vito might provide without the need to install a website again.


saeedvaziry:
What is the use case for this? it can end up with wrong scenarios. Like changing Laravel to Wordpress

dyanakiev:
Actually yes this could cause some havoc, it might be better to not allow it.

Issue #298, opened 2024-10-02
Laravel Commands


saeedvaziry:
From: https://vitodeploy.featurebase.app/p/laravel-commands

Create a new section in the sites to support Laravel artisan commands like:

- Cache
- Up/Down
- Queue (failed, restart ...)
- Custom commands
- Write down here if you think more commands are needed.

saeedvaziry:
Blocked by #297

link2dawood:
1. Cache Management
Command: **php artisan cache:clear**
Description: Clears all the cache data for the application.
Use Case: Useful when data is outdated and needs to be refreshed.

Related Commands:
**php artisan config:cache** – Caches the configuration files.
**php artisan route:cache** – Caches the routes for better performance.
**php artisan view:clear** – Clears compiled views.

2. Maintenance Mode (Up/Down)
Command: **php artisan down**
Description: Puts the application into maintenance mode, preventing users from accessing it while updates or changes are being made.
Use Case: Ideal for major updates or database migrations.
Command: **php artisan up**
Description: Brings the application back online after maintenance.

3. Queue Management
Command: **php artisan queue:restart**
Description: Restarts queue workers to apply changes to the application or queue configurations.
Use Case: Use when you update code that is being processed by workers.

Command: **php artisan queue:failed**
Description: Shows a list of all failed jobs in the queue.
Command: **php artisan queue:flush**
Description: Deletes all failed queue jobs.
Command: **php artisan queue:work**
Description: Starts processing jobs in the queue.

4. Custom Commands
Command: **php artisan make:command CustomCommandName**
Description: Creates a new custom Artisan command that can be used for specific application tasks.
Use Case: Useful for adding tailored functionality to your Laravel application, such as scheduled tasks or batch processing.

5. Database Commands
Command: **php artisan migrate**
Description: Runs all outstanding migrations.
Command: **php artisan migrate:rollback**
Description: Rolls back the last database migration.

6. Other Useful Commands
Command: **php artisan serve**
Description: Starts a local development server.
Command: **php artisan tinker**
Description: Opens an interactive shell for running Laravel/PHP code in real-time.

alinuxist:
Do you support Inertia SSR?

CalumChamberlain:
@alinuxist 
> Do you support Inertia SSR?

I've just been looking for the same thing. I think I've got it working by going to sites -> queues -> new queue and using the following command

php /home/vito/site name/artisan inertia:start-ssr

[...]

PR #519, opened 2025-02-28
feat: Implement Site Commands (#298)

dyanakiev:
- Introduced a Commands widget/table for Site view, allowing users to create, edit, delete, and execute commands.
- Each Site Type now has a predefined set of commands inserted upon site creation.
- A migration script ensures commands are created for existing sites.
- Implemented necessary policies for command management.

CleanShot 2025-03-01 at 00 57 01


saeedvaziry:
Made some changes. I will add some tests as well

saeedvaziry:
Also need to add the feature to the API as well

PR #520, opened 2025-03-01
feat: Add support for wildcard subdomains in domain validation regex

dyanakiev:
The previous regex did not support wildcard domains (`*.example.com`), as it required domain labels to start with an alphanumeric character. This update modifies the regex to allow `*` as a valid standalone subdomain label at any level.

Changes:
- Allowed `*` as a valid subdomain at any depth.
- Ensured `*` remains a standalone label (e.g., `*-sub.example.com` is still invalid).
- Maintained existing domain validation rules for non-wildcard domains.

Now supports:
✅ `*.example.com`
✅ `*.sub.example.com`
✅ `sub.*.example.com`
✅ `*.sub.sub2.example.com`

Still prevents:
❌ `*-sub.example.com`
❌ `sub.-*.example.com`

This improves compatibility with wildcard DNS records while maintaining proper domain validation constraints.


dyanakiev:
Example Use Case:

A multi-tenant SaaS platform allows customers to configure wildcard subdomains (*.example.com) for dynamic tenant assignment (e.g., client1.example.com).

saeedvaziry:
@dyanakiev I would appreciate some tests to cover all the proposed cases

saeedvaziry:
I think this will be problematic. Vito uses domain to create folders for the websites. This will break it.

PR #521, opened 2025-03-01
feat: Allow editing of site domain in Site details

dyanakiev:
Added functionality to display and edit the site domain in the site details section. This allows users to update their domain directly within the app, making it easier to manage domain changes.

CleanShot 2025-03-01 at 11 09 07

### DIFF
--- a/app/Actions/Site/UpdateDomain.php
+++ b/app/Actions/Site/UpdateDomain.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace App\Actions\Site;
+
+use App\Models\Site;
+use App\SSH\Services\Webserver\Webserver;
+use App\ValidationRules\DomainRule;
+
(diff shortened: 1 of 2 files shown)


saeedvaziry:
I don't think this would be that much easy.

We have the site directory as the domain, in some cases we get the site path from its domain.

this can raise some instability issues and I would avoid it tbh

dyanakiev:
> I don't think this would be that much easy.
> 
> We have the site directory as the domain, in some cases we get the site path from its domain.
> 
> this can raise some instability issues and I would avoid it tbh

Upon further consideration, I recognize the potential pitfalls associated with altering the domain name. The domain name is currently utilized for various purposes, including the PHP-FPM socket and nginx vhosts, which may not be ideal for such a change.

PR #522, opened 2025-03-01
feat: Improve Site folder naming

dyanakiev:
Previously, when creating a Site, the generated folder name included the full domain with its TLD (e.g., `example.com`). This had a few drawbacks:
- **Unnecessary Complexity**: Including the TLD made folder names longer and added redundant information.
- **Filesystem Safety**: Some TLDs contain special characters that aren't ideal for folder naming, like `.бг`, `.рф`. These may cause issues due to non-Latin characters.
- **Readability & Cleanliness**: Shorter, cleaner folder names improve organization and make it easier to navigate directories.

### **Before (Old Folder Naming)**
- `example.com` → `example.com`
- `sub.example.com` → `sub.example.com`
- `my-site.dev.co.uk` → `my-site.dev.co.uk`

### **Now (New Folder Naming)**
- `example.com` → `example_a8f3d2`
- `sub.example.com` → `sub_example_b9e5c4`
- `my-site.dev.co.uk` → `my_site_dev_a9c2f1`

### **Why This is Better**
- **Removes the TLD** to keep folder names cleaner and more structured.
- **Ensures safe characters** using `Str::slug()`.
- **Adds a short unique identifier** to prevent accidental conflicts while keeping names readable.
- **Enforces lowercase formatting** for consistency.
- **Better for Domain Changes**: When a website changes its domain (e.g., `example.com` → `example.net`), the old `.tld` won't remain in the folder name, preventing confusion and keeping directories clean.

This change makes folder naming cleaner, more readable, and more filesystem-friendly. 🚀

### DIFF
--- a/app/Actions/Site/CreateSite.php
+++ b/app/Actions/Site/CreateSite.php
@@ -14,6 +14,7 @@ use App\Notifications\SiteInstallationSucceed;
 use App\ValidationRules\DomainRule;
 use Exception;
 use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Str;
 use Illuminate\Validation\Rule;
 use Illuminate\Validation\ValidationException;
 
@@ -30,7 +31,7 @@ class CreateSite
                 'domain' => $input['domain'],
                 'aliases' => $input['aliases'] ?? [],
                 'user' => $user,
-                'path' => '/home/'.$user.'/'.$input['domain'],
+                'path' => '/home/'.$user.'/'.self::generateUniqueAppFolderName($input['domain']),
                 'status' => SiteStatus::INSTALLING,
             ]);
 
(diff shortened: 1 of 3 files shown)


slaFFik:
I'm against this change, as in a lot of places (scripts, cron commands, etc.) you need to manually type the folder where the site is located, and now you can't do that from memory, as you will need to check the generated folder name every single time.

dyanakiev:
I got the idea from serverpilot, which I've used before.

As an alternative maybe asking for a Application name would be better?
Runcloud example:
CleanShot 2025-03-02 at 05 00 03

saeedvaziry:
> example.com → example_a8f3d2

This really doesn't look nice

AiAe:
I do agree this should be fixed in some shape or form, even if its not this solution, there is still probably few more that can be implemented.
Like get rid of the tld from the folder name, use the git repo name or give the option to enter one ourselves?

And since we know this will break existing projects, can't we just make migration to fix previously made projects, so they keep working without issues?

RichardAnderson:
Can we not just strip out everything thats not safe?  I personally like having the full domain as the path, it's much easier to run scripts against (like someone else mentioned).

dyanakiev:
> Can we not just strip out everything thats not safe? I personally like having the full domain as the path, it's much easier to run scripts against (like someone else mentioned).

I agree with your idea. We could add a field called 'Site folder name' below the Domain field. This field would automatically populate when the user types the domain. Then, we can use str()->slug() when populating the new field to ensure that the site path is always safe and consistent. The user will also be able to edit the site folder name as needed.

Before implementing this, we should do a full run through all the files to standardize the use of the app path or app id instead of relying on places on the domain name to resolve the site path. This will make Vito more reliable and consistent. 

I think both this and #521 have a common problem.

PR #523, opened 2025-03-01
fix: add missing backticks in Blade for MySQL/MariaDB database names with hyphens

dyanakiev:
Previously, database names with hyphens (e.g., my-database) failed due to missing backticks.

Before:
The CREATE DATABASE command failed for database names containing hyphens (-) because MySQL/MariaDB requires backticks around such names, but they were missing in the Blade template.

After:
Backticks (`) are now correctly added in the Blade template, ensuring MySQL/MariaDB properly interprets database names with hyphens.

Why:
MySQL/MariaDB does not allow hyphens in database names unless they are enclosed in backticks. The Blade template was missing these backticks, causing syntax errors. Adding them ensures the command executes successfully. 🚀


saeedvaziry:
This is intentional to avoid some querying issues like:
`using hyphens can cause issues when querying (e.g., SELECT * FROM my-db-name.table_name will cause an error without backticks`

dyanakiev:
> This is intentional to avoid some querying issues like: `using hyphens can cause issues when querying (e.g., SELECT * FROM my-db-name.table_name will cause an error without backticks`

The same issue arises if I attempt to use reserved keywords without enclosing them in backticks.
```
CREATE TABLE order (id INT, name VARCHAR(50))
> 1064 - You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'order (id INT, name VARCHAR(50))' at line 1
> Query Time: 0.069s
```

Most frameworks automatically wrap identifiers in backticks (or equivalent) to prevent such syntax issues, not just for database names but also for reserved keywords in queries.

PR #532, opened 2025-03-02
fix: Sail default container name to match docker-compose.yml

dyanakiev:
By default, Sail looks for the `laravel.test` container when running `sail artisan`, but the `docker-compose.yml` in this project defines the service as `app`. This caused `sail artisan` and other Sail commands to fail locally.

Updated the `APP_SERVICE` environment variable in `sail` to default to `app` instead of `laravel.test`, ensuring consistency with the service name in `docker-compose.yml`.

### DIFF
--- a/sail
+++ b/sail
@@ -122,7 +122,7 @@ fi
 
 # Define environment variables...
 export APP_PORT=${APP_PORT:-80}
-export APP_SERVICE=${APP_SERVICE:-"laravel.test"}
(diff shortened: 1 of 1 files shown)


saeedvaziry:
Merging this, but I recommend using the env variables.

dyanakiev:
> Merging this, but I recommend using the env variables.

Env variables are the best way to go, but since the default sail service in docker-compose.yml has been changed, it should be okay to change it in the sail executable as well.

Issue #535, opened 2025-03-02
Bug Report: An issue in the migration of "Database collations #489" could potentially break Vito installations if executed.


dyanakiev:
**Describe the bug**
When executing the `php artisan migrate` command, the `syncDatabases` function attempts to establish an SSH connection to each server.
However, an issue arises when a server is offline, resulting in an exception message indicating "Cannot connect to x.x.x.x:22. Error 65. No route to host." This exception causes the migration to fail, preventing it from being marked as complete. Furthermore, attempting to retry the migration is not feasible, as new columns have already been created during the initial execution.

**To Reproduce**
1. Stop a server or temporary set wrong IP
2. Run `php artisan migrate`
3. See error

**Screenshots**
![Image](https://github.com/user-attachments/assets/73cf16c7-410e-4dbd-9d8d-be22b1cbf18a)

**Additional context**
https://github.com/vitodeploy/vito/blob/5a12ed76bb1bfd342f68d9fbf21740bd09587d58/database/migrations/2025_02_15_120213_update_databases_table.php#L35

The problematic PR: Database collations #489

dyanakiev:
A potential solution could involve incorporating a try-catch block around the problematic line: $this->syncDatabases(false) in the migration file. I read somewhere there will be added a button to manually refresh the databases if i'm not mistaken so they could run that if they decided to turn back online their server

PR #537, opened 2025-03-02
Add database and database users sync

### LINKED ISSUE #374: Database Sync
Sometimes it is possible that you create a database or database user using the command line in the server.

In order to sync such information to vito as well, we need to put a button in the database (header actions) to fetch all users and databases and their links and update them in vito's database.

It is also important to prevent syncing the system reserved databases :

- mysql
- information_schema
- peformance_schema
- sys
- postgres
- template0
- template1

saeedvaziry:
Closes #374


dyanakiev:
Closes #535 as well

PR #550, opened 2025-03-16
Enable opcache by default for better performance.

talhawish:
OPCache can significantly improve performance by caching precompiled script bytecode, reducing the need for repeated parsing and compiling of PHP code on each request. Having the ability to toggle this feature would provide more flexibility for users to optimize performance based on their specific use cases.

### DIFF
--- a/docker/php.ini
+++ b/docker/php.ini
@@ -1,3 +1,12 @@
 [PHP]
 post_max_size = 100M
 upload_max_filesize = 100M
+
+
+; ----- OPCache Settings -----
+
+; Enable OPCache for performance improvement
+opcache.enable=1
+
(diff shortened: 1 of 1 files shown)


saeedvaziry:
Agree. But this PR is going to enable OPCache on Vito instance itself not the websites you deploy using Vito :)

talhawish:
I was on github having a loom, searched php.ini , updated and committed
from browser, my bad ,  I haven’t yet installed, I just started a project
which I’ll be deploying soon. 🥹

saeedvaziry:
Thanks for your contribution,
I recommend trying it out and then a PR is welcome to add this

PR #551, opened 2025-03-17
feat(api): add endpoint to update site aliases

J4Wx:
Allows the updating of a sites aliases using a PUT request. The request requires one property on the body, "aliases" - which is an array of strings.

I have not ran `scribe:generate` (yet) - the entire scribe directory gets rewritten and I'm not familiar enough with scribe to know if this is the correct behaviour.

I intend to continue adding more API functionality, but this is something that has a particular priority for me currently (hence PRing this alone first).

### DIFF
--- a/app/Http/Controllers/API/SiteController.php
+++ b/app/Http/Controllers/API/SiteController.php
@@ -125,7 +125,7 @@ class SiteController extends Controller
 
         $this->validateRoute($project, $server, $site);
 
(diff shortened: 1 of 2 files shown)


J4Wx:
Test added!

It's possible that I'm just an idiot, but it might be worth adding to the contributing docs (or maybe just the readme?) that you need to manually create `storage/database-test.sqlite` before running the tests.

It only took 30 seconds for me to realise, but might help some people.

PR #552, opened 2025-03-18
feat(redirects): add redirects to sites

### LINKED ISSUE #299: Site Redirects
Add 301 and 302 redirects to sites

J4Wx:
Resolves #299 

* Supports index, creating and deleting redirects via UI and API.
* Supports status 301, 302, 307, 308.
* Updates the vHosts.
* Created tests to cover API and UI use.

### DIFF
--- a/app/Actions/Redirect/CreateRedirect.php
+++ b/app/Actions/Redirect/CreateRedirect.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace App\Actions\Redirect;
+
+use App\Models\Redirect;
+use App\Models\Site;
+use Illuminate\Validation\Rule;
+
+class CreateRedirect
+{
+    public function create(Site $site, array $input): Redirect
+    {
+        $redirect = new Redirect();
+
+        $redirect->site_id = $site->id;
+        $redirect->from = $input['from'];
+        $redirect->to = $input['to'];
+        $redirect->mode = $input['mode'];
+        $redirect->status = 'created'; // This field isn't currently used.
+
+        $redirect->save();
+
+        /** @var Service $service */
+        $service = $site->server->webserver();
+
+        /** @var Webserver $webserver */
+        $webserver = $service->handler();
+        $webserver->updateVHost($site);
+
+        return $redirect;
+    }
+
+    public static function rules(Site $site): array
+    {
+        return [
+            'from' => ['required', 'string', 'max:255', 'not_regex:/^http(s)?:\/\//', Rule::unique('redirects', 'from')->where('site_id', $site->id)],
+            'to' => ['required', 'url:http,https'],
+            'mode' => ['required', 'integer', 'in:' . implode(',', [
+                301,
+                302,
+                307,
+                308,
+            ])],
+        ];
+    }
+}
--- a/app/Actions/Redirect/DeleteRedirect.php
+++ b/app/Actions/Redirect/DeleteRedirect.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace App\Actions\Redirect;
+
+use App\Models\Redirect;
+use App\Models\Site;
+
+class DeleteRedirect
+{
+    public function delete(Site $site, Redirect $redirect): void
+    {
+        $redirect->delete();
+
+        /** @var Service $service */
+        $service = $site->server->webserver();
+
+        /** @var Webserver $webserver */
+        $webserver = $service->handler();
+        $webserver->updateVHost($site);
+    }
+    
+    public static function rules(): array
+    {
+        return [];
+    }
+}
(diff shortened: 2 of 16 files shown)


J4Wx:
I am fixing 👀

dyanakiev:
Hey @J4Wx, I noticed that there doesn't seem to be a database migration file. Is there something I'm missing?

J4Wx:
@dyanakiev 

Weirdly... the project already contains a redirect migration that wasn't used at all.

It already contained everything I needed, I assume it was created at some point early in the projects life but never got implemented.

If there's something I'm missing and that table was used historically then I'll create a second table for backwards compatibility if needed.

dyanakiev:
@J4Wx,

I was looking at the tables but didn't see it right away. It seems like there's no existing usage of the redirects table.

I think someone started working on that and left some files in there.

I noticed another mention of redirects in this file: [resources/views/ssh/services/webserver/nginx/update-redirects.blade.php](https://github.com/vitodeploy/vito/blob/2.x/resources/views/ssh/services/webserver/nginx/update-redirects.blade.php). The idea there was to store the redirects in a separate config file, but your solution works just as well.

In discussions, I had an idea to store any extra config files in a directory so everything in that directory will be included in the vhost. You can check out this discussion: https://github.com/vitodeploy/vito/discussions/530. That way, it'll be clearer.
`/etc/nginx/sites/custom-configs/${SITE_ID}/*.conf`

J4Wx:
@dyanakiev 

I think that supporting additional config files like that would be very useful - I'm working on that change now.

J4Wx:
@dyanakiev 

vHosts files now include two "include" lines:
```
include /etc/nginx/conf.d/extension-configs/{{ $site->id }}/*.conf;
include /etc/nginx/conf.d/custom-configs/{{ $site->id }}/*.conf;
```

`/etc/nginx/conf.d/extension-configs/{{ $site->id }}/*.conf` is used for additional automatically generated configs, like `/etc/nginx/conf.d/extension-configs/{{ $site->id }}/redirects.conf`. These things are transient and will be changed programatically.

`include /etc/nginx/conf.d/custom-configs/{{ $site->id }}/*.conf;` is for user created extensions to the vHosts file.

---

Redirects are now generated within the redirects.conf file specified above.

The webserver contract (and Nginx.php of course) now have an additional method, createOrUpdateAdditionalConfig, with the signature: `createOrUpdateAdditionalConfig(Site $site, string $configName, ?string $configContent = null, bool $restart = true)`.
* If you specify a configContent then it uses that content and places the file in custom-configs. (This currently isn't used, but can be used in future if we add the ability to create these custom configs within Vito).
* If you don't specify a configContent, then the content is generated from the respective view, and place in extension-configs.

Moving forward, additional configuration can be moved inside extension-configs instead of using the vHosts file directly, reducing how often it'll be regenerated (I'm thinking SSL configuration at the very least). Even if added additional folders for outside of the `server` directive - for things like the loadbalancer's `upstream` and force_ssl sections.

Is this the kind of thing you had in mind?

dyanakiev:
@J4Wx Yes, that’s exactly what I had in mind, great job! 🎉

saeedvaziry:
Created a temp url from `/test` to `httpx://..../test2` and didn't forward it.

Also facing errors when deleting

CleanShot 2025-03-29 at 18 07 07@2x

J4Wx:
@saeedvaziry can you show me what ended up being in your vHost and in the `/etc/nginx/conf.d/extension-configs/{{ $site->id }}/redirects.conf`, please?

saeedvaziry:
```
vito@test:/home/vito$ cat /etc/nginx/conf.d/extension-configs/43/redirects.conf
location /test.php {
return 302 https://test.vitodeploy.com;
}
```

app/SSH/Services/Webserver/Nginx.php
```diff
@@ -96,13 +103,52 @@ public function createVHost(Site $site): void
...
+
+        $this->ensureConfigPathsExist($site);
+    }
+
+    /**
+     * @param  string  $configName  The name of the config file to create or update, without the .conf extension. If $configContent is null, the config file will be created with the contents of the view file.
+     *
+     * @throws SSHError|ViewNotFoundException
+     */
+    public function createOrUpdateAdditionalConfig(Site $site, string $configName, ?string $configContent = null, bool $restart = true): void
```
saeedvaziry, line 115: I won't go with this approach. I'd rather follow the simple rules. Open for extension and close for modification. I appreciate the vision but I believe this hard to understand and maintain.

saeedvaziry:
This looks ok but I would slightly change the way this PR uses nginx to redirect.

Right now its based on `location` but looks like using `rewrite` is better practice

J4Wx:
I would agree that rewrite should be used if we were supporting more complex cases, but for simple redirects I'm not aware of any reason to choose rewrite over `location { return }`.

I'm happy to be wrong about that, though.

saeedvaziry:
gonna add some docs now

PR #560, opened 2025-03-22
feat: Add site cloning functionality

dyanakiev:
CleanShot 2025-03-22 at 10 38 52
# Site Duplication Feature

## Description
This PR adds the ability to duplicate existing sites within the same server. Users can create exact copies of sites with new domains and optionally different branches, making it easier to create staging/testing environments or replicate site configurations.

## Features
- New `DuplicateSite` action for handling site duplication logic
- Site duplication modal in site management UI
- Support for:
  - Custom domain
  - Domain aliases
  - Branch selection (for sites with source control)
- Complete copying of site files
- Comprehensive test coverage

## API Changes
Added new endpoint for site duplication:

### Duplicate Site
`POST /api/projects/{project}/servers/{server}/sites/{site}/duplicate`

#### Authentication
- Requires authentication token
- Requires 'write' ability

#### Request Parameters
| Parameter | Type | Required | Description |
|-----------|------|----------|-------------|
| domain | string | Yes | New domain for the duplicated site |
| aliases | array | No | Array of domain aliases |
| branch | string | No | Branch name (for sites with source control) |

#### Example Request
```json
{
    "domain": "staging.example.com",
    "aliases": ["www.staging.example.com"],
    "branch": "develop"
}


saeedvaziry:
I think would be good to have an option for isolated users. Currently this PR clones the site under the same user as the cloned site which might not be the use case for isolated users. The idea of isolated users is to have one user per site.

dyanakiev:
I added the option to enter username if wanted.

saeedvaziry:
@dyanakiev why didn't use the existing Action (CreateSite) and created a new Action? I would go with re-using the existing code than duplicating it

saeedvaziry:
Also by cloning I thought we will create a new site with the same configs not literally copying the existing site. by copying it will also get some unwanted files like stuff in the storage and envs and etc

dyanakiev:
> Also by cloning I thought we will create a new site with the same configs not literally copying the existing site. by copying it will also get some unwanted files like stuff in the storage and envs and etc

Thats the idea, to clone all existing files as well. Could be used for wordpress staging sites or similar situations.

ChrisToxz:
Can't we make it an option? I see use cases for both.
A duplicate for staging, testing, reproducing etc. Including files and config, but I would say without .env, because that would manually changes anyway (E.G. Site URL & DB)

But also a fresh copy, as Saeed suggest, to have a clean start, without storage.

saeedvaziry:
Looks a bit complex now. offering this feature might be useful for WP sites but maybe not Laravel sites due to env, storage folder, cache and ...

HoroTW:
Just wanted to chime in because this functionality would be useful for me (a new WordPress page would be copied from a "template" page).

eriteric:
Yes the use case for wordpress is much simpler. 
This feature is commonly found in panel hosts. It makes a big difference when you can one-click a staging environment. 
+1

dyanakiev:
I think copying the entire folder and its contents is fine even for Laravel applications including the .env, if anyone wants can always delete the files he needs.

Maybe i can add "advanced" settings where there will be choice to skip certain files or folders.

HoroTW:
After looking into it a bit more....
I started writing something that does exactly the tasks I need.... I think my use cases are just too different from vitodeploys features...
E.g. in the WordPress case it is not enough to just copy it, there are db (and file) settings configured for the URI, so one would need additional steps... (e.g. doable with wp-cli in the WordPress case) Still a big fan of vitodeploy, even if it might not exactly match my needs :wink:

viezel:
@saeedvaziry can we add this feature to the v3 roadmap, but with your suggestions - meaning it's more focused on the Laravel usecase where we do not clone all the site files, but rather clone the setup in an isolatation fashion?

saeedvaziry:
@viezel https://github.com/vitodeploy/vito/discussions/568#discussioncomment-13047876

medanielsantos:
We believe this feature is especially effective for WordPress-based projects, where quickly replicating environments can significantly speed up development and testing.

saeedvaziry:
Closing as outdated

PR #567, opened 2025-03-24
fix(site/workers): add-all-ssh-users-to-worker-edit

J4Wx:
The worker create action currently allows you to select any SSH user across the server: https://github.com/vitodeploy/vito/blob/2.x/app/Web/Pages/Servers/Workers/Actions/Create.php#L27

But the worker edit action only allows you to choose "vito" or "root".
https://github.com/vitodeploy/vito/blob/2.x/app/Web/Pages/Servers/Workers/Widgets/WorkersList.php#L150

This results in a broken looking form whenever editing a worker, and needing to delete and recreate a worker if you need to change to a user that isn't "root" or "vito".


saeedvaziry:
How about this?

```
->options(
                   $site ? 
                       array_combine($site->getSshUsers(), $site->getSshUsers()) : 
                       array_combine($server->getSshUsers(), $server->getSshUsers())
                ),
```

for Edit and Create worker

PR #569, opened 2025-03-25
feat(ssl): support multiple certificates per site

J4Wx:
The rationale behind my desire for this change is explained here:
https://github.com/vitodeploy/vito/discussions/557

So, what does this change do?

* You can now have multiple SSL certificates activated per site.
* Each certificate can be assigned to a set of domains (the main domain and each of the aliases are options).
* Custom certificates can be edited so that you manage the domains assigned to use it.
* Activating a new certificate for a domain/alias (or editing a custom domain to add a domain/alias) will remove that domain/alias from the assignment for other certificates.
* The `server {}` block template has been moved from the usual vHost template so that it can be included programmatically with different variables. Each certificate will include it's own block, with the correct values set. A default non SSL block will be included too, supporting all domains.

What are the downsides?

* Ultimately it can result in a really long vHosts file. I considered the alternative of using something like `map $ssl_host_name $sslKey{{$site_id}}` to dynamically load the SSL config, but that would require making each SSL cert readable by the configured user for that site (since dynamic loading would happen after nginx de-escalates it's permissions from root). This issue can be mitigated in future with the change from https://github.com/vitodeploy/vito/pull/552 - more sections of the config can be dynamically loaded and thus minimise change still.
 
Migration?

* It doesn't change the location of certs or private keys, so existing vHosts files will continue to work as normal.
* It leverages the already used `domains` field on the SSL table. So if an existing vHost is regenerated then it'll continue to work as normal.


saeedvaziry:
There has to be at least one domain provided when generating SSL. otherwise it will fail:
```
Saving debug log to /var/log/letsencrypt/letsencrypt.log

Plugins selected: Authenticator nginx, Installer nginx

Account registered.

Missing command line flag or config entry for this setting:
Which names would you like to activate HTTPS for?
We recommend selecting either all domains, or all domains in a VirtualHost/server block.
test.vitodeploy.com?

(You can set this with the --domains flag)

Ask for help or search for solutions at https://community.letsencrypt.org. See the logfile /var/log/letsencrypt/letsencrypt.log or re-run Certbot with -v for more details.
```

saeedvaziry:
Closing as outdated

PR #607, opened 2025-06-06
fix the branch required validation rule for create-site on first loading

oliverquynh:
On the create-site screen first loading, although the branch field is set as "main", the actual form value is empty (missing). 

Screenshot 2025-06-06 104756

That's why the branch required validation rule does not pass. It displays "The branch field is required." after submitting.

https://github.com/user-attachments/assets/35b21105-aeac-49c5-84e0-0e37adb129cc

I've set "main" as the default branch in the form.

### DIFF
--- a/resources/js/pages/sites/components/create-site.tsx
+++ b/resources/js/pages/sites/components/create-site.tsx
@@ -38,6 +38,7 @@ export default function CreateSite({ server, children }: { server?: Server; chil
     aliases: [],
     php_version: '',
     source_control: '',
+    branch: 'main',
     user: '',
   });
 


saeedvaziry:
Good catch. I think we need to change the approach here.  Dynamic fields will be for plugins that want new field without touching the frontend. so it should work frontend agnostic.

We can set in the DynamicField component to set the field on the initial load first.

oliverquynh:
@saeedvaziry I just realize one more FE problem. The create-site form isn't scrolled vertically. If there are more fields or error messages, they'll take lots of spaces and the buttons are pushed down, like coming out of the screen. We won't see and can't click them.

I think it's not limited to the create-site form. It's the general side (right) form problem.

saeedvaziry:
Now I see. same for create server. I guess adding a Scrollarea should solve the issue

saeedvaziry:
I've already resolved this in a local branch. will push soon

PR #614, opened 2025-06-17
Update CreateServer action

oliverquynh:
I'm using Windows for PHP development at my current company. I'm using Herd to serve HTTP requests, and docker for running queue workers. If SSH key files are created by docker containers, everything will work fine. But, if they are created via HTTP requests (Windows environment), it won't work. I have to manually fix private key files permission.

Therefore, I'd like to make an update here. Instead of calling `$this->server->provider()->create()` right in a HTTP request, we can call it inside a background job. SSH keys will be created by queue workers in Debian docker containers. We won't need to care about file permission on Windows anymore.

Here is a quick review:

```php
// Storing data into database in HTTP requests as usual
$this->server->save();
$this->createFirewallRules();
$this->createServices();

dispatch(function () {
    // Create SSH files and store in our filesystem
    // Call third-party API to add SSH public key to cloud provider
    // Call third-party API to create a real server from cloud provider
    $this->server->provider()->create();

    // Wait for 30 seconds before checking if SSH connection is available
    Sleep::for(30)->seconds();

    // Do server installation
    app(InstallServer::class)->run($this->server);
})
```

It's just a small change but does not break current behavior, but support development on Windows.

### DIFF
--- a/app/Actions/Server/CreateServer.php
+++ b/app/Actions/Server/CreateServer.php
@@ -15,6 +15,7 @@ use Exception;
 use Illuminate\Database\Query\Builder;
 use Illuminate\Support\Facades\Log;
 use Illuminate\Support\Facades\Validator;
+use Illuminate\Support\Sleep;
 use Illuminate\Support\Str;
 use Illuminate\Validation\Rule;
 use Illuminate\Validation\ValidationException;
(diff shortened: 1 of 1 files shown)


app/Actions/Server/CreateServer.php
```diff
@@ -162,9 +162,9 @@ private static function providerRules(array $input): array
         return $server->provider()->createRules($input);
     }
 
-    public function createFirewallRules(Server $server): void
+    public function createFirewallRules(): void
```
oliverquynh, line 165: Both `createFirewallRules` and `createServices` creates resources for the newly create server, so we can use `$this->server` in `createFirewallRules` instead of passing `$this->server` itself as a parameter.

saeedvaziry:
@seriquynh the reason we have it in the HTTP call to make sure that the server is being created on the cloud provider when using any of them. if for example Hetzner throws an error reaching the limit or duplicated keys or whatever, we would notice it right before we start the installation and not wait for the job to throw an error and then we check the logs.

PR #615, opened 2025-06-19
Restrict Site User Management to Administrators Only

algethamy:
This PR modifies the site user management functionality to restrict the ability to add users to sites to administrators only. Previously, any user associated with a site's project could add other users to the site, which could lead to security concerns and unintended access permissions.

## Security Benefits
- Enhances security by ensuring only administrators can control site access
- Prevents regular users from granting access to other users
- Maintains the principle of least privilege


saeedvaziry:
This will make user management more complex. also if a user can have access to the server, technically can access to all its sites. limiting on Vito doesn't make sense.

PR #686, opened 2025-08-05
fix: .env.sail missing APP_SERVICE

arzola:
Without this value any `sail` command will throw `service "laravel.test" is not running`

This is helpful when running vito on sail for local development.


saeedvaziry:
`vito` would be better no?

arzola:
I like that but I think that will require to update the docker-compose right? I can do that

Something like...
```
services:
    vito:
```

saeedvaziry:
ah I thought APP_SERVICE=vito but that wouldnt work

PR #536, opened 2025-03-02
feat: Console Auto-Completion and Keyboard Shortcuts

dyanakiev:
preview

## Description  
This PR introduces several new features to the console interface, significantly improving its functionality and user experience. The changes include:

### New Keyboard Shortcuts  
Implement new, familiar shell shortcuts to streamline command editing:
- **`Ctrl+U`** - Clear line before the cursor  
- **`Ctrl+K`** - Clear line after the cursor  
- **`Ctrl+W`** - Delete the previous word  
- **`Ctrl+A`** - Move the cursor to the start of the line  
- **`Ctrl+E`** - Move the cursor to the end of the line  
- **`Alt+B`** - Move backward by one word  
- **`Alt+F`** - Move forward by one word  
- **`Ctrl+C`** - Cancel the current command  
- **`Ctrl+L`** - Clear the output  

### Advanced Tab Completion  
- Introduces dynamic tab-based auto-completion for both commands and directory paths.  
- Users can cycle through suggestions using `Tab` and `Shift+Tab`, and the `Enter` key applies the selected suggestion.  
- Suggestions are filtered in real time based on user input.
- Adds logic to fetch and display both command and directory suggestions separately.  
- Dynamically fetches directory contents when using the `cd` command, ensuring relevant suggestions.
- Integrates a scrollable dropdown for displaying suggestions, with improved focus handling during navigation.  
- Ensures a smooth and intuitive user experience when interacting with auto-completion.

These new features align the console's behavior more closely with standard terminal interfaces, enhancing overall usability.

PR #693, opened 2025-08-06
Add console tab autocomplete functionality

DiogoDuart3:
- Add backend endpoint for directory listing and file suggestions
- Implement frontend tab completion with arrow key navigation
- Add loading indicator during autocomplete requests
- Support for both absolute and relative path completion
- Clean dropdown UI for file/folder suggestions
- Tab completion works with commands like 'ls ', 'cd ', etc.

### DIFF
--- a/app/Http/Controllers/ConsoleController.php
+++ b/app/Http/Controllers/ConsoleController.php
@@ -90,4 +90,57 @@ class ConsoleController extends Controller
 
         return response()->json(['status' => 'ok']);
     }
+
+    #[Post('/autocomplete', name: 'console.autocomplete')]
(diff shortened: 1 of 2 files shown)


saeedvaziry:
what should we expect from this? for example cd only works for the current level and doesn't go further.

CleanShot 2025-08-07 at 10 50 43

AiAe:
Why not expend this PR to add what it was done in https://github.com/vitodeploy/vito/pull/536 
Or at least to take some of the ideas?

saeedvaziry:
Closing as stale. Feel free to open another PR

PR #694, opened 2025-08-06
feat: add command history functionality to console component

DiogoDuart3:
This pull request adds command history navigation to the console input, allowing users to cycle through previously entered commands using the up and down arrow keys. This enhancement improves user experience by making it easier to recall and reuse commands.

**User experience improvements:**

* Added state variables `commandHistory` and `historyIndex` to track the list of entered commands and the current position in the history.
* Updated command execution logic to trim the command, add it to history, and reset the history index after each command run.
* Implemented keyboard navigation for command history: pressing the up arrow populates the input with the previous command, and pressing the down arrow moves forward in the history or clears the input if at the end.


saeedvaziry:
We could store the history in the local storage to be useful even after reloading the page. WDYT?

PR #697, opened 2025-08-08
Add and remove versions services

JurgenThijs:
- Removed the versions for PHP below 8.3 as they don't provide security updates anymore 
- Removed the versions for NodeJS 16 and 18 as no security updates are provided anymore 
- Added Nodejs 24
- Added MariaDB version 11.8


saeedvaziry:
I won't be deleting those old versions as some might need for their lagecy apps. its up to user decide to what to use and we're not here to force them. latest versions always recommended to use

saeedvaziry:
Closing this. feel free to open it for mariadb 11.8

PR #701, opened 2025-08-09
fix: Composer path for plugin installation

RichardAnderson:
The composer path was being included in the PATH, but also included once again in the command, removing this from the command resolves any issues where the path to the composer binary is non-standard (for example, has a space within one of the folders).

This allows plugins to be installed when using Herd or other non-standard setups.

### DIFF
--- a/app/Plugins/Plugins.php
+++ b/app/Plugins/Plugins.php
@@ -58,7 +58,7 @@ class Plugins
         $output = $result->output();
 
         if ($result->failed()) {
-            throw new Exception($output);
+            throw new Exception($result->errorOutput());
         }
 
         $output .= $this->load();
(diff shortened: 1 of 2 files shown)


saeedvaziry:
How about this for composer_path?

```
$composer = new \Illuminate\Support\Composer(new \Illuminate\Filesystem\Filesystem);
return implode(' ', $composer->findComposer());
```

RichardAnderson:
> How about this for composer_path?
> 
> ```
> $composer = new \Illuminate\Support\Composer(new \Illuminate\Filesystem\Filesystem);
> return implode(' ', $composer->findComposer());
> ```

This always returns "composer" in my tests, and therefore, fails, as it does not include the full path.

saeedvaziry:
> This always returns "composer" in my tests, and therefore, fails, as it does not include the full path.

@RichardAnderson what you have [here](https://github.com/vitodeploy/vito/pull/701/files#diff-11f16095ffd7b4a724da6c587d87142af1c0d03ee46caa8c6eae557259db7281R100) is also `composer` no?

PR #710, opened 2025-08-12
fix Gitlab auto deployment

mahavir-atliq:


saeedvaziry:
I am not sure about this. I need to check it. also if this is how gitlab sends the webhooks then we should just simply delete that function from the Gitlab.php to let it to use the one in the parent class which is what you have in this PR

PR #731, opened 2025-08-19
feat: plugin installation scripts

arzola:
### Why? ⚠️

I think Plugins should be able to perform expensive "one time" operations during plugin's **installation** and **uninstallation** process, creating files, configuring database, etc... this way we won't pollute the `boot` method with extra checks that should have to run on every kernel request, current Plugins architecture won't run composer scripts as Regular Laravel Packages that are added in the main composer.json file so we need to run those scripts manually after loading them.

With this change Plugins will be able to execute things like:

```json
{
    "name": "arzola/extra-php-extensions",
    "description": "A Vito plugin to fetch and make available PHP extensions.",
    "type": "library",
    "license": "MIT",
    "version": "1.0.0",
    "authors": [
        {
            "name": "Oscar Arzola",
            "email": "os@icode.mx"
        }
    ],
    "scripts": {
        "test": "vendor/bin/phpunit",
         "post-package-install": [
           "php artisan cache:clear",
           "php artisan php-extensions:fetch"
         ],
        "pre-package-uninstall": [
           "php artisan cache:clear",
           "php artisan php-extensions:uninstall"
         ]
    }
}
```

**Plugin lifecycle and script execution enhancements:**

* Added execution of `post-package-install` scripts after plugin installation and `pre-package-uninstall` scripts before plugin removal, using the plugin's `composer.json` file. These scripts are tracked with a flag file in `storage/plugins/.installed/` to prevent duplicate execution. [[1]](diffhunk://#diff-11f16095ffd7b4a724da6c587d87142af1c0d03ee46caa8c6eae557259db7281R103-R104) [[2]](diffhunk://#diff-11f16095ffd7b4a724da6c587d87142af1c0d03ee46caa8c6eae557259db7281R124-R131) [[3]](diffhunk://#diff-11f16095ffd7b4a724da6c587d87142af1c0d03ee46caa8c6eae557259db7281R165-R205)
* Improved plugin uninstall logic to remove the flag file after successful uninstallation, ensuring accurate tracking of installed plugins.

The `.installed` directory it's a good way to track which packages were actually installed that way we don't re-run scripts in case we call manually `php artisan plugins:load`

### Additional Bug fix

Pasted_Image_2025-08-19__5_32 PM

I found that entering package's URL with a trailing slash will "install the plugin in a wrong directory" (without vendor namespace)

* Fixed the extraction of vendor and name from plugin URLs by trimming trailing slashes, improving compatibility with various URL formats.

### Added tests for Plugins

* Added unit tests in `tests/Unit/Plugins/PluginsTest.php` to cover plugin installation, loading, uninstalling, lifecycle script execution, and error handling, ensuring reliability and correctness of the plugin system.


arzola:
Btw I can update docs for the events we recently added and if this goes through I can add that too

saeedvaziry:
This is looking good. however tests are failing locally
image

arzola:
I'll take a look :)

saeedvaziry:
Does it consider octane plugin already installed or it assumes there are no plugins? cause in my local and all default vito installations octane is already there

saeedvaziry:
Yes. I see after I remove the default laravel octane plugin tests are passing

arzola:
I think maybe I can improve the test to not consider other plugins when running tests

arzola:
@saeedvaziry I updated the tests to move existing plugins temporarily during tests

saeedvaziry:
Thanks @arzola for the contribution

PR #754, opened 2025-08-22
Use site path as directory option for worker supervisor conf

rygilles:
Discussion https://github.com/vitodeploy/vito/discussions/753


saeedvaziry:
@rygilles with this we don't need to put a full path to the queue workers command anymore?

rygilles:
@saeedvaziry Yes, it's like doing `cd /home/vito/site && ` before running the command.
e.g. It allows you not to have to do `php /home/vito/site/artisan horizon` but `php artisan horizon` instead

saeedvaziry:
@rygilles are you still working on it? hit the `Ready for review` if all done

PR #770, opened 2025-08-27
Add timezone support to user profile updates

DiogoDuart3:
- Introduced a new TimezoneRule for validating timezones.
- Updated UserProfileInformation and ProfileController to handle timezone input.
- Enhanced the profile update form to include a timezone selection.
- Added tests to ensure timezone can be updated and invalid timezones are rejected.

### DIFF
--- a/app/Actions/User/UpdateUserProfileInformation.php
+++ b/app/Actions/User/UpdateUserProfileInformation.php
@@ -3,6 +3,7 @@
 namespace App\Actions\User;
 
 use App\Models\User;
+use App\ValidationRules\TimezoneRule;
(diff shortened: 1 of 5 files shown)


saeedvaziry:
That timezone field is stale and not being used at all AFAIK. it was being used in vito 2

on v3, We are using moment timezone in `resources/js/components/date-time.tsx`

DiogoDuart3:
Won’t it be used for display porpuses? When a user sees a table and there’s some datetime there

saeedvaziry:
No it wouldn't

saeedvaziry:
Thanks for the effort but I am closing this as it won't change anything.

PR #780, opened 2025-08-31
fix: correct spelling of "connections" in headings across multiple pages

azadshaikh:
correct spelling of "connections" in headings across multiple pages


resources/js/pages/server-settings/index.tsx
```diff
@@ -218,7 +218,7 @@ export default function Databases() {
           </CardHeader>
           <CardContent>
             <div className="space-y-2 p-4">
-              <p>please note that this action is irreversible and will delete all data associated with the server.</p>
+              <p>This action will transfer the server to another project. All associated data will remain intact.</p>
```
saeedvaziry, line 221: Good one!

PR #798, opened 2025-09-04
enable plugins to add server features

flowan:
- These changes will introduce server features (it kind of works the same as site features)
- Added a simple server feature plugin called MOTD (Message of the day)
- Textarea added to the dynamic field (added it to be able to create a field to set a custom MOTD)


saeedvaziry:
Thanks for the contribution, looking good.

I would remove the className as there is no guaranty that the class you enter will be generated during the build (tailwind + vite)

putting this on hold in favor of #800 as some major refactors will happen to the plugins system.

saeedvaziry:
The MOTD doesn't seem to be working. I enabled as position start , end or even replace but on ssh login it doesn't show

saeedvaziry:
@flowan I appreciate the PR however I don't think it is necessary to include the MOTD feature. dropped that part as it can be a plugin anyway. will merge the PR

flowan:
> @flowan I appreciate the PR however I don't think it is necessary to include the MOTD feature. dropped that part as it can be a plugin anyway. will merge the PR

Hey! No problem, i was thinking of adding it as a plugin. So no hard feelings :)

PR #807, opened 2025-09-06
remove legacy plugins

saeedvaziry:


RichardAnderson:
As part of install/upgrade we are deleting all legacy plugins, do we still need the LoadLegacyPlugins command?

saeedvaziry:
> As part of install/upgrade we are deleting all legacy plugins, do we still need the LoadLegacyPlugins command?

@RichardAnderson we need to keep it at least for the next release. because the update script already has the load:plugins and it will fail. in fact we need to bring back the plugins:load command instead of load legacy plugins command.

saeedvaziry:
So since people don't just upgrade to the next release and there might be release jumps. we will need to keep that command even for a while

RichardAnderson (approved):

saeedvaziry:
I am going to delete all other commands related to legacy plugins

PR #808, opened 2025-09-06
Repo & Branch Search (Github Only)

RichardAnderson:
## Repo & Branch Search (with fallback) for Github
Provides the ability for repos and branches to be pulled from GitHub automatically, along with a filter for searching for a given repo.

https://github.com/user-attachments/assets/a842a6c2-e8a5-490b-9cf2-84ef31d34d59

Only implemented the functionality for GitHub, others will fall back to the default of returning an empty array, which will prompt the user to enter a repo/branch manually.


saeedvaziry:
Proposal for after this PR:

Would be nice if we finally introduce some settings/configurations for Vito itself so user can change some stuff.

We can start by exposing this as settings, if for example dynamic repo/branch was on we do this otherwise we just show normal text inputs

saeedvaziry (approved):

PR #809, opened 2025-09-06
[Fix] Mocks Github actions in tests to avoid rate limiting

RichardAnderson:
Removes the retrieve of versions and download of the plugin from Github to avoid rate limiting when running the plugins locally or via CICD in quick succession.


RichardAnderson:
Test failing on build, but not on my local machine.  Will investigate tomorrow.

app/Actions/Plugins/Github/DownloadRelease.php
```diff
@@ -6,7 +6,7 @@
 use Exception;
 use Illuminate\Support\Facades\File;
 
-final readonly class DownloadRelease
```
saeedvaziry, line 9: I didn't know how to tell you don't use final at all haha

PR #837, opened 2025-09-21
Add domain editing functionality

dyanakiev:
I decided to attempt another version of my previous PR in v2 for domain editing.

**Features:**
- Domain editing modal with safety warnings
- Comprehensive validation including domain uniqueness, site readiness checks, and modern deployment blocking
- Site-level worker recreation to ensure processes run from correct directories after domain changes
- Webserver configuration updates for both Nginx and Caddy with proper file moving and symlink handling

CleanShot 2025-09-21 at 17 58 49

### DIFF
--- a/app/Actions/Site/UpdateDomain.php
+++ b/app/Actions/Site/UpdateDomain.php
@@ -1,0 +1,181 @@
+<?php
+
+namespace App\Actions\Site;
+
+use App\Models\Service;
+use App\Models\Site;
+use App\Services\Webserver\Webserver;
+use App\ValidationRules\DomainRule;
+use Illuminate\Support\Facades\Validator;
+use Illuminate\Validation\Rule;
+
+class UpdateDomain
+{
+    /**
+     * @param  array<string, mixed>  $input
+     */
+    public function update(Site $site, array $input): void
+    {
+        $this->validate($site, $input);
+
+        $oldDomain = $site->domain;
+        $newDomain = $input['domain'];
+
+        // If domain hasn't changed, do nothing
+        if ($oldDomain === $newDomain) {
+            return;
+        }
+
+        // Check if site is ready for domain change
+        if (! $site->isReady()) {
+            throw new \Exception('Cannot change domain while site is not ready.');
+        }
+
+        // Check if modern deployment is enabled - it would break the system
+        if ($site->modernDeploymentEnabled()) {
+            throw new \Exception('Cannot change domain while modern deployment is enabled. Please disable modern deployment first, change the domain, then re-enable it.');
+        }
+
+        // Update the domain and path in the database
+        $site->domain = $newDomain;
+        $site->path = '/home/'.$site->user.'/'.$newDomain;
+        $site->save();
+
+        // Move the physical directory on the server
+        $this->moveSiteDirectory($site, $oldDomain, $newDomain);
+
+        // Update webserver configuration files
+        $this->updateWebserverConfig($site, $oldDomain, $newDomain);
+
+        // Recreate site level workers to pick up the new site path
+        $this->recreateWorkers($site);
+    }
+
+    protected function moveSiteDirectory(Site $site, string $oldDomain, string $newDomain): void
+    {
+        $oldPath = '/home/'.$site->user.'/'.$oldDomain;
+        $newPath = '/home/'.$site->user.'/'.$newDomain;
+
+        // Move the directory from old path to new path
+        $site->server->ssh()->exec(
+            "sudo mv {$oldPath} {$newPath}",
+            'move-site-directory',
+            $site->id
+        );
+    }
+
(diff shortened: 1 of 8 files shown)


app/Actions/Site/UpdateDomain.php
```diff
@@ -0,0 +1,181 @@
...
+
+        // Move and update the virtual host file with new domain
+        $this->moveAndUpdateVirtualHost($site, $oldDomain, $newDomain, $webserver);
+    }
+
+    protected function moveAndUpdateVirtualHost(Site $site, string $oldDomain, string $newDomain, Webserver $webserver): void
+    {
+        $webserverId = $webserver::id();
+
+        if ($webserverId === 'nginx') {
```
saeedvaziry, line 83: This logic can't live inside the actions. Actions don't know what webserver server has. They need to agnosticly let the handler of the webserver to handle it

app/Actions/Site/UpdateDomain.php
```diff
@@ -0,0 +1,181 @@
...
+
+        // Check if site is ready for domain change
+        if (! $site->isReady()) {
+            throw new \Exception('Cannot change domain while site is not ready.');
+        }
+
+        // Check if modern deployment is enabled - it would break the system
+        if ($site->modernDeploymentEnabled()) {
+            throw new \Exception('Cannot change domain while modern deployment is enabled. Please disable modern deployment first, change the domain, then re-enable it.');
+        }
```
saeedvaziry, line 37: You can throw ValidationException::withMessages to so user sees the error under the domain field

app/Actions/Site/UpdateDomain.php
```diff
@@ -0,0 +1,181 @@
...
+    {
+        $this->validate($site, $input);
+
+        $oldDomain = $site->domain;
+        $newDomain = $input['domain'];
+
+        // If domain hasn't changed, do nothing
+        if ($oldDomain === $newDomain) {
+            return;
+        }
```
saeedvaziry, line 27: not necessary if you remove the line `->ignore...` from the validation

app/Actions/Site/UpdateDomain.php
```diff
@@ -0,0 +1,181 @@
...
+                    $worker->site->path,
+                    $worker->site_id
+                );
+            } catch (\Throwable $e) {
+                // Log the error but continue with other workers
+                \Illuminate\Support\Facades\Log::error("Failed to restart worker {$worker->id}: ".$e->getMessage());
+            }
+        }
+    }
+
```
saeedvaziry, line 160: I'd say lets just not let domain to be changed if there are workers 🤔

app/Actions/Site/UpdateDomain.php
```diff
@@ -0,0 +1,181 @@
...
+                    ->ignore($site->id),
+            ],
+        ])->validate();
+
+        // Additional validation to ensure site is ready
+        if (! $site->isReady()) {
+            $validator = validator([], []);
+            $validator->errors()->add('domain', 'Cannot change domain while site is not ready.');
+            throw new \Illuminate\Validation\ValidationException($validator);
+        }
```
saeedvaziry, line 179: why double validation?

app/Actions/Site/UpdateDomain.php
dyanakiev: Yea i agree, will refactor

app/Actions/Site/UpdateDomain.php
dyanakiev: I don't know really. I don't like that we are also preventing even when modern deployment is enabled, but I'm not that familiar with the logic behind modern deployments, so I didn't write code to handle that. A better solution would be to not change the site folder name at all, but that will need to be rethought generally how folder names are done, which I guess is not on the agenda. If you want, I will add prevention of changing the domain when there are workers and remove the recreate code. I don't see it as an issue that we are recreating it, though?

app/Actions/Site/UpdateDomain.php
dyanakiev: Yeah, I forgot I moved that. Will get rid of it in the update method. Is that good?

app/Actions/Site/UpdateDomain.php
saeedvaziry: Actually I don't think this will cause a problem at all if the modern deployment was enabled. we can test it of course.

Workers have logs and they have ids in the log files and etc... thats why maintaining this part will be difficult for us. I would even say lets not touch the workers and tell user that they need to change themselves

fatihalp:
@dyanakiev these kind of critical actions; I would delete and add new domain.

In my experience, it's important to consider what happens when problems arise in these types of operations (such as when the move doesn't work). Therefore, I believe that using the same structure as existing scenarios (such as delete and add) will be more predictable and maintainable. Just my humble advice.

saeedvaziry:
Closing as out dated

PR #838, opened 2025-09-22
Add Vito self-backup system

dyanakiev:
This PR implements a self-backup feature for Vito that was previously requested by users who needed to backup their Vito instance data and configuration.

### Why this is needed:
- Users have been requesting a way to backup their Vito installations https://github.com/vitodeploy/vito/discussions/792
- Current backup systems only handle user databases, not Vito itself
- Self-backup ensures Vito configuration and data can be restored if needed

### What's included:
- Backup scheduling with cron-like frequencies (hourly, daily, weekly, monthly)
- Support for S3, Dropbox, and FTP storage providers
- Manual backup triggers and file management
- Cleanup of old backups based on retention settings

The system automatically backs up Vito's database and configuration files to the selected storage provider, giving users peace of mind that their Vito setup can be restored.

CleanShot 2025-09-22 at 12 37 33

### DIFF
--- a/app/Actions/VitoBackup/BackupOperations.php
+++ b/app/Actions/VitoBackup/BackupOperations.php
@@ -1,0 +1,109 @@
+<?php
+
+namespace App\Actions\VitoBackup;
+
+use Exception;
+use Illuminate\Support\Facades\File;
+use Illuminate\Support\Facades\Storage;
+use ZipArchive;
+
+trait BackupOperations
+{
+    /**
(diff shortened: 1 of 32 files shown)


saeedvaziry:
I really like the table design of this feature but I would go with the one we already have just to keep things consistent

saeedvaziry:
And actions can be dropdown like other places in the app

saeedvaziry:
Closing as out dated

PR #845, opened 2025-09-29
If there are no webservers, scribe:generate fails

oralunal:
In an empty Vito installment without any created webserver, api documentation generation fails.

Call to a member function webserver() on null
```
  at app/Models/Site.php:455
    451▕ 
    452▕     public function webserver(): Webserver
    453▕     {
    454▕         /** @var Service $webserver */
  ➜ 455▕         $webserver = $this->server->webserver();
    456▕ 
    457▕         /** @var Webserver $handler */
    458▕         $handler = $webserver->handler();
    459▕ 

      +2 vendor frames 

  3   app/Http/Resources/SiteResource.php:27
      Illuminate\Http\Resources\Json\JsonResource::__call("webserver", [])
      +1 vendor frames 

  5   [internal]:0
      Illuminate\Support\HigherOrderCollectionProxy::{closure:Illuminate\Support\HigherOrderCollectionProxy::__call():63}(Object(App\Http\Resources\SiteResource))
```


saeedvaziry:
Why do we need this? the release action working fine already.

oralunal:
I've installed vito locally to measure resource usage for the "scribe:generate" command and i realized this. I assume that this is a bug, maybe?

saeedvaziry:
Well try to run it after seeding database using demo mode = on then it should be fine.

I don't want to continue with scribe. there is a lot of files.

Will drop scribe anyways in favor of something more simple

oralunal:
Okay, so, this pr can be closed now :)

PR #872, opened 2025-10-08
Remove unnecessary index drop statements from migration files

thanhtrungit97:
This PR removes the following line from the migration file:

```php
DB::statement('DROP INDEX IF EXISTS plugins_is_enabled_priority_index');
```

because the index plugins_is_enabled_priority_index does not exist in the current schema, and the statement causes a syntax error when using the MySQL driver, since MySQL does not support the DROP INDEX IF EXISTS syntax.

### DIFF
--- a/database/migrations/2025_09_21_113237_add_site_id_to_cron_jobs_table.php
+++ b/database/migrations/2025_09_21_113237_add_site_id_to_cron_jobs_table.php
@@ -12,7 +12,9 @@ return new class extends Migration
      */
     public function up(): void
     {
-        if (Schema::hasTable('plugins')) {
+        if (DB::getDriverName() === 'sqlite') {
+            DB::statement('DROP INDEX IF EXISTS plugins_is_enabled_priority_index');
+        } elseif (Schema::hasTable('plugins')) {
             $indexes = DB::select("SHOW INDEXES FROM plugins WHERE Key_name = 'plugins_is_enabled_priority_index'");
             if (!empty($indexes)) {
                 DB::statement('DROP INDEX plugins_is_enabled_priority_index ON plugins');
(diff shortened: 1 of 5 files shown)


dyanakiev:
That's too extreme. Perhaps implement a solution from here:
https://gist.github.com/Razoxane/3bc74900b4eb5c983eb0927fa13b95f5?permalink_comment_id=5535833#gistcomment-5535833
```php
<?php

namespace App\Traits;

use Illuminate\Database\Schema\Blueprint;
use Illuminate\Support\Facades\Schema;

trait MigrationIndex
{
    public function dropIndexIfExist(string $tableName, string $indexName): void
    {
        $indexes = Schema::getIndexes($tableName);

        foreach ($indexes as $index) {
            if ($index['name'] === $indexName) {
                Schema::table($tableName, static function (Blueprint $table) use ($indexName) {
                    $table->dropIndex($indexName);
                });
            }
        }
    }
}
```

thanhtrungit97:
@dyanakiev  But we are sure that index did not exist before.

saeedvaziry:
Why is this important? I put them on every migration after the bug intentionally because not everyone running the same vito version. the wrong index name breaks every migration.

I don't see any harm letting it live there

thanhtrungit97:
@saeedvaziry Have you tested it on DB_CONNECTION=mysql ?

saeedvaziry:
if mysql failing we can update those migration to also run on mysql. but I would avoid deleting them.

thanhtrungit97:
@saeedvaziry 
Thank you, I understand and updated the PR

saeedvaziry:
@thanhtrungit97 my point being. you can just wrap the existing statements into if db driver is sqlite and leave mysql.

saeedvaziry:
if I am not mistaken

```php
if (DB::getDriverName() === 'sqlite') {
   // drop those indexes
}
```

thanhtrungit97:
@saeedvaziry Okay! updated

PR #886, opened 2025-10-17
Add Domains and DNS management

### LINKED ISSUE #884: Manage Domains via DNS Providers
## DNS Providers

Add DNS Providers integrations in the Settings so user can connect to the DNS providers like Cloudflare. This will provide an interface that can manage Domain's DNS records mainly.

## Domains

Add a Domains section to the main menu that will use the DNS Provider connections to add domains from those providers to Vito and uses the DNSProvider interface to manage the most common NS records through Vito.

saeedvaziry:
closes #884


saeedvaziry:
Also closes #887

PR #943, opened 2025-12-04
Add Forgejo as SourceControlProvider

Elhebert:
This adds support for repositories hosted on [Codeberg](https://codeberg.org) by adding [Forgejo](https://codeberg.org/forgejo/forgejo) as a possible Source Control Provider.

While the API looks to be the same, Forgejo is a hard-fork of Gitea so we can't use the Gitea SourceControlProvider for repositories hosted on a Forgejo instance.

### DIFF
--- a/app/Providers/SourceControlServiceProvider.php
+++ b/app/Providers/SourceControlServiceProvider.php
@@ -7,6 +7,7 @@ use App\DTOs\DynamicForm;
 use App\Plugins\RegisterSourceControl;
 use App\SourceControlProviders\Bitbucket;
 use App\SourceControlProviders\BitbucketV2;
+use App\SourceControlProviders\Forgejo;
 use App\SourceControlProviders\Gitea;
 use App\SourceControlProviders\Github;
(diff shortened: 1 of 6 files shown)


saeedvaziry:
Please make a Plugin instead

Elhebert:
What is the reason behind Gitea being a first party service provider then? Not sure I understand when something should be a plugin and when it should be core.

saeedvaziry:
what is a problem adding it as a plugin?

My main reasons not wanting this in the core:

- I never used it and never even heard of it. Gitea is more popular
- Letting to core means more maintenance for me while I can focus on features that we don't have yet

Elhebert:
It's not a problem to add it as a plugin. Just wanted to know since Gitea was present on top of the usual 3 that we see everywhere.

I would say, however, that the plugin system isn't great at the moment. It relies 100% on GitHub. Testing is also really complicated since plugin requires to be integrated within the Vito codebase to run, on their own they can't be ran.

PR #950, opened 2025-12-09
Improve Site Selection Logic for Server-Specific Validation

natanbiz:
This PR refactors the site selection logic to ensure that only the sites valid for the currently selected server are displayed and available for use. When the server changes, any previously selected sites are automatically cleared, and the stored site data is updated accordingly.


saeedvaziry:
Seems you need a new PR :)

PR #960, opened 2025-12-15
Add fail-on-error to Scripts

emredipi:
## Problem

Scripts executed in `runScript()` would continue running even when a command failed. For example, if a migration failed or git pull errored, the script wouldn't stop and would keep executing subsequent commands. This could lead to incorrect deployment states.

## Solution

Added `set -e` and `set -o pipefail` at the beginning of scripts. Now when any command fails, the script stops and the SSH exec method properly throws an error.

- `set -e`: Script stops when a command returns a non-zero exit code
- `set -o pipefail`: Enables error checking in pipes as well

### DIFF
--- a/app/SSH/OS/OS.php
+++ b/app/SSH/OS/OS.php
@@ -237,7 +237,9 @@ class OS
         if ($serverLog instanceof ServerLog) {
             $ssh->setLog($serverLog);
         }
-        $command = "shopt -s expand_aliases\n";
+        $command = "set -e\n";
+        $command .= "set -o pipefail\n";
+        $command .= "shopt -s expand_aliases\n";
         if ($aliases !== null && $aliases !== []) {
             foreach ($aliases as $key => $alias) {
                 $command .= "alias $key=$alias\n";


saeedvaziry:
I don't think its a good idea to merge this to V3. it will bring breaking changes to some scripts that even throwing error in them is considered fine and script has to finish executing.

We can have this on `4.x` instead however on `4.x` we might need this in the SSH.php for all commands

emredipi:
Hi @saeedvaziry,

I understand the breaking change concern, and you’re right about that.

That said, the current behavior has a real downside: when a command fails, we don’t get any signal or warning at all. From my perspective, it’s hard to imagine someone intentionally relying on completely silent failures. Even if a script finishes successfully, it may leave the system in an incorrect or partially deployed state without anyone noticing.

So I think there is a clear problem here. This particular solution might not be the final answer for V3, but the underlying issue is definitely something we should address in one way or another.

emredipi:
Also worth noting: even if something goes wrong during a deployment, the process still returns a success response, which makes the failure completely invisible from the outside.
Ekran Resmi 2025-12-16 13 58 13

saeedvaziry:
My concern is not other user's scripts will break. everyone who uses deployment scripts and the main scripts they can already put `set -e` on top of the script and it will fail if error happens.

My main concern is Vito itself and all the places Vito relies on script getting finished no matter if it fails. I cannot test every feature of vito on a real server with different conditions to make sure this works so thats why I'd like to have such feature on the new major version that we will anyway test every part and have a beta period.

saeedvaziry:
merged to 4.x, will optimize it when get to work with the general fail-on-error topic

emredipi:
Thank you @saeedvaziry 🙏

PR #970, opened 2025-12-24
Add password()/passwordWithToggle() DynamicFields

0aveRyan:
## Adds
* `password()` and `passwordWithToggle()` input types for `DynamicFields`.
* Styled and sized like `text()` with password inputs to let Site Types declare a form field with visually-hidden values.

password

passwordWithToggle


saeedvaziry:
I think for future we need to refactor dynamic-field component to smaller files

PR #973, opened 2025-12-26
feat: make create database dialog select default charset and collation

### LINKED ISSUE #939: Default database configuration
### Discussed in https://github.com/vitodeploy/vito/discussions/760

<sup>Originally posted by **marky291** August 23, 2025</sup>
As a user when creating a mysql database, I do not want to have to think about what collation and charset to use, I should see default values with optional change if needed:

'MySQL Server has a server character set and a server collation. By default, these are utf8mb4 and utf8mb4_0900_ai_ci , but they can be set explicitly at server startup on the command line or in an option file and changed at runtime.'

https://dev.mysql.com/doc/refman/8.4/en/charset-server.html#:~:text=MySQL%20Server%20has%20a%20server,file%20and%20changed%20at%20runtime.</div>

JmzTaylor:
Matches defaults in MySQL 8+.  Doesn't throw error if charset or collection doesn't exist.

Fixes #939 #760

### DIFF
--- a/resources/js/pages/databases/components/create-database.tsx
+++ b/resources/js/pages/databases/components/create-database.tsx
@@ -64,12 +64,12 @@ export default function CreateDatabase({
 
   // Auto-load collations when modal opens with a default charset
   useEffect(() => {
-    if (open && defaultCharset && charsets.includes(defaultCharset) && collations.length === 0) {
-      axios.get(route('databases.collations', { server: server, charset: defaultCharset })).then((response) => {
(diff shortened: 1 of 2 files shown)


saeedvaziry:
Thanks for the PR. How this will work for Postgresql?

JmzTaylor:
It doesn't change anything for Postgresql.  It will show up exactly as before.

saeedvaziry:
No I mean we should also select default values for Postgres as well

JmzTaylor:
Updated the commit.  This will gracefully fallback to current behavior if the default charset or collation doesn't exist.

PR #976, opened 2025-12-26
add set -e to deployment scripts

JmzTaylor:
Allows any failure to stop deployment

Closes discussion https://github.com/vitodeploy/vito/discussions/851


saeedvaziry:
Discussed here #960 

We're going to tackle this properly by adding it on core SSH helper in v4

Issue #781, opened 2025-09-01
Clarification on base64 encoding in SSH command


algethamy:
Hi there,

I hope you’re doing well.

I was reviewing the following line in the SSH helper:

https://github.com/vitodeploy/vito/blob/c338b688f92f2032dcbf5c390c8a5b9efbd70316/app/Helpers/SSH.php#L124

I noticed that the command is being base64 encoded before execution. Could you please explain the rationale behind encoding the command this way?
Is it mainly for escaping or safety purposes when running sudo su -c?

Thanks in advance!

saeedvaziry:
Its being used for primary single quote and double quotes issues in the commands which happens mostly to isolated users.

Since those isolated users don't have ssh (because they are isolated) vito ssh using vito user and runs command with sudo for that isolated users

PR #998, opened 2026-01-20
Adjust SSH heredoc wrapper

algethamy:
trying to fix this #781 

What changed
	•	Switched to using bash with a quoted heredoc (<<'EOF') when executing commands as another user.
	•	Updated the related test case to assert the new heredoc-based command structure.
	•	Removed unnecessary base64 encoding/decoding from the execution flow.

Why this change
	•	Improves readability of executed commands.
	•	Avoids complex shell quoting and nested bash -c calls.
	•	Prevents unintended variable expansion by using a quoted heredoc.
	•	Makes the execution logic more robust and easier to reason about in tests.

Notes
	•	No functional behavior change in command execution logic, only the execution method.
	•	Tests were updated accordingly to reflect the new expected command format.

### DIFF
--- a/app/Helpers/SSH.php
+++ b/app/Helpers/SSH.php
@@ -170,8 +170,11 @@ class SSH
 
         try {
             if ($this->asUser !== null && $this->asUser !== '' && $this->asUser !== '0') {
-                $command = base64_encode((string) $command);
-                $command = "sudo su - {$this->asUser} -c 'bash -c \"echo {$command} | base64 -d | bash\"'";
+                $command = <<<BASH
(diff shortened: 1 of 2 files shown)


saeedvaziry:
To test this I literally need to test the entire vito. I don't have time for it unfortunately at the moment

algethamy:
Hey @saeedvaziry 👋
Many existing test cases in the repo already use this function with the new approach and all of them pass, so the behavior is well-covered and stable.  ￼

Can you review / merge when you get a chance? Thanks! 👍

PR #1000, opened 2026-01-20
fix: service versions in desc order

victorelec14:
Return service versions in descending order.

before:
Captura de pantalla 2026-01-20 222024

after:
image

Thx :)


saeedvaziry:
lets push for v4

PR #1017, opened 2026-02-01
Prevent cross-site SSL inheritance

### LINKED ISSUE #1010: [Bug]: New sites without SSL inherit another site’s HTTPS vhost (isolation issue)
### Installation Type *

VPS

### Vito Version *

3.2

### Describe the Bug *

When creating a new site in VitoDeploy without SSL enabled, HTTPS traffic for that domain is routed to an existing site that already has SSL configured. This results in the wrong site’s content being served and breaks application installers such as WordPress.

This appears to be a design issue in the site-creation workflow rather than a misconfiguration.

### Steps to Reproduce *

Have an existing site on the server with SSL enabled (e.g. site-a.com)
Create a new site via VitoDeploy (e.g. site-b.com)
Select an application type (WordPress, Laravel, etc.)
Complete database creation
Visit https://site-b.com

### Expected Behavior *

HTTPS traffic for a site without SSL should:
return an error, or
be handled by a neutral default SSL vhost (sink)

### Logs *

Can't really provide a log for this one.

### Screenshots

_No response_

### Additional Context

To prevent HTTPS requests for newly created sites from being routed to an existing site with SSL enabled, I added a neutral “default SSL sink” vhost in Nginx.

This ensures that when a domain does not yet have SSL configured, HTTPS requests fail closed (connection dropped) instead of serving another site’s content. Once SSL is enabled for the site, its own vhost automatically takes precedence.

This was done as a one-time hardening step to avoid cross-site routing during site creation and application installs (e.g. WordPress), and does not affect existing SSL-enabled sites.

saeedvaziry:
closes #1010

### DIFF
--- a/app/Services/Webserver/Nginx.php
+++ b/app/Services/Webserver/Nginx.php
@@ -43,6 +43,22 @@ class Nginx extends AbstractWebserver
             'root'
         );
 
+        $this->service->server->ssh()->exec(
+            view('ssh.services.webserver.nginx.create-default-ssl'),
+            'create-default-ssl'
+        );
+
+        $this->service->server->ssh()->write(
+            '/etc/nginx/sites-available/000-default-ssl',
+            view('ssh.services.webserver.nginx.default-ssl-vhost'),
+            'root'
+        );
+
+        $this->service->server->ssh()->exec(
+            'sudo ln -sf /etc/nginx/sites-available/000-default-ssl /etc/nginx/sites-enabled/000-default-ssl',
+            'enable-default-ssl'
+        );
+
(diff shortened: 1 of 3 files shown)


dyanakiev:
Hey @saeedvaziry i don't see a way to apply this to existing servers. Unless manually.

saeedvaziry:
Yeah old servers need to handle it manually or just generate SSL for every site.

I haven't released this yet. If I can find a way for existing servers, I'll push

dyanakiev:
Running it as migration seems like possiblity

saeedvaziry:
More in favor of doing it when a new site being created than running it blindly on all servers

PR #1059, opened 2026-03-22
[Feat] WebSockets Improvements

RichardAnderson:
This pull request introduces improved WebSocket connection handling, security and user feedback in both the backend (PHP) and frontend (React/TypeScript) parts of the application. The most significant changes include a more robust reconnection strategy for WebSocket connections, new UI elements to indicate connection status and allow manual reconnection, and a simplification of connection management on the server side.

# Features / Improvements

- Improved tracking of web socket connections on the UI
- Automatic retries limited to 3 initial attempts, then backs off for 30 seconds before trying again.
- Ability to manually reconnect by clicking the icon in the header
- Persistent WebSocket connection: Moved WebSocket connection management from React hooks into a Zustand store (socket-store.ts) so the connection survives Inertia page navigations instead of tearing down and reconnecting on every route change                                                
- Dedicated broadcast secret: Replaced usage of APP_KEY as the WebSocket broadcast bearer token with a dedicated WS_BROADCAST_SECRET env var. Falls back to APP_KEY with a deprecation warning for existing deployments for production. (Potential of leaking APP_KEY via logging or monitoring services, for example, if someone has configured Nightwatch, etc.) 
- Origin header validation: Added isOriginAllowed() check during WebSocket handshake to prevent cross-site WebSocket hijacking. Configurable via WS_ALLOWED_ORIGINS env var, falls back to APP_URL                         
- Subscribe rate limiting: Added sliding window rate limiter (20 requests/60s per user) on subscribe messages to prevent DB query abuse via allProjects()->exists() lookups 

# Notes

## Dev Setup

- New .env variable required for allowed origins: `WS_ALLOWED_ORIGINS="${APP_URL}"`
- `APP_URL` must be set to the domain vito is being served on locally (what you are accessing the URL by)
- Set `WS_URL` if this differs from your web socket implementation (for example, locally WS_URL might be `http://localhost`


config/core.php
```diff
@@ -21,6 +21,9 @@
      */
     'ws_host' => env('WS_HOST', '127.0.0.1'),
     'ws_port' => env('WS_PORT', '8085'),
+    'ws_broadcast_secret' => env('WS_BROADCAST_SECRET', env('APP_KEY')),
+    'ws_broadcast_secret_set' => ! empty(env('WS_BROADCAST_SECRET')),
```
saeedvaziry, line 25: A dynamic config depends on another?

saeedvaziry (approved):

config/core.php
RichardAnderson: As a fallback, yeah, allows for keeping it as app_key in development, but shows a warning:

CleanShot 2026-03-23 at 08 13 12

saeedvaziry: I mean it could be just `if (!config('core. ws_broadcast_secret'))`

PR #1062, opened 2026-03-23
[Fix] Site Domains Fixes

RichardAnderson:
SSL certificate filtering and validation:

* Updated the `Ssl` model's `activeServerLevel` scope to exclude certificates of type `CSR` and require both `certificate_path` and `pk_path` to be present, ensuring only valid, usable server-level SSLs are returned.
* Added the `SslType` enum import to the `Ssl` model to support the above filtering.

SSL expiry warnings:

* Improved the logic for SSL expiry warnings in `SiteResource`: now checks that `expires_at` is set before comparing, and safely handles the case where no expiry date is found when determining the earliest expiry.

API schema cleanup:

* Removed the unused `is_active` property from the SSL OpenAPI schema, reflecting changes in how active SSLs are determined.

### DIFF
--- a/app/Models/Ssl.php
+++ b/app/Models/Ssl.php
@@ -3,6 +3,7 @@
 namespace App\Models;
 
 use App\Enums\SslStatus;
+use App\Enums\SslType;
 use Carbon\Carbon;
 use Database\Factories\SslFactory;
 use Illuminate\Database\Eloquent\Builder;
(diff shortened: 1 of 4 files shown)


app/Models/Ssl.php
saeedvaziry, line 117: Please add a todo for query optimization

saeedvaziry (approved):

PR #1075, opened 2026-04-22
[FIX] v3.x: Match synced MySQL users by (username, host) to avoid collapsing multi-host accounts

### LINKED ISSUE #1074: [Bug] v3.x: SyncDatabaseUsers corrupts `databases` column when multiple MySQL users share a username across different hosts
## Installation Type *

VPS

## Vito Version *

3.21.2

## Describe the Bug *

`SyncDatabaseUsers::updateUsers()` looks up existing rows by `username` only (`app/Actions/Database/SyncDatabaseUsers.php:33`). MySQL/MariaDB treat `'user'@'localhost'` and `'user'@'127.0.0.1'` as separate accounts, so the sync collapses them onto a single Vito row: one host keeps getting its `databases` overwritten on every sync, the second host variant is never inserted, and the stored `host` ends up not matching the grants shown above it. No error is thrown, the drift is silent.

## Steps to Reproduce *

On a Vito-managed MySQL/MariaDB server, create two accounts that share a username but differ on host (common once an app connects via TCP instead of the unix socket):

```sql
CREATE USER 'app'@'localhost' IDENTIFIED BY 'secret';
CREATE USER 'app'@'127.0.0.1' IDENTIFIED BY 'secret';
GRANT ALL ON appdb.* TO 'app'@'localhost';
GRANT ALL ON appdb.* TO 'app'@'127.0.0.1';
GRANT ALL ON devdb.* TO 'app'@'127.0.0.1';
```

Then click Sync in the UI (`PATCH /servers/{server}/database-users/sync`) and inspect `database_users` rows for that server.

## Expected Behavior *

One Vito row per `(username, host)` pair, each carrying the grants that actually apply to that account.

## Logs *

No log output. The bug only shows up as drift between `SHOW GRANTS` on the server and what `database_users` contains in Vito's own DB.

## Screenshots

N/A.

## Additional Context

`CreateDatabaseUser`, `UpdateDatabaseUser`, and `DeleteDatabaseUser` already scope by `(username, host)`; this sync path is the outlier. PostgreSQL is unaffected because its `get-users-list` view returns an empty host (roles are host-agnostic), so any fix needs to keep the username-only branch when host is empty. Fix attempt in #1075.

erhanurgun:
Closes #1074

`SyncDatabaseUsers` was looking up existing rows by username only, so on a server with both `'app'@'localhost'` and `'app'@'127.0.0.1'` one row kept being overwritten on every sync and the second one was never inserted. The Vito UI then ended up showing a `databases` list that didn't match the host above it.

Fix is one extra `where('host', ...)` when the handler returns a non-empty host. Postgres is left on the username-only path because its `get-users-list` view returns `''` for host (roles are host-agnostic) and an unconditional composite filter would insert a duplicate `(rolname, '')` row on every sync.

```diff
  /** @var ?DatabaseUser $databaseUser */
- $databaseUser = $server->databaseUsers()
-     ->where('username', $user[0])
-     ->first();
+ $query = $server->databaseUsers()->where('username', $user[0]);
+ if ($user[1] !== '') {
+     $query->where('host', $user[1]);
+ }
+ $databaseUser = $query->first();
```

`vendor/bin/phpunit tests/Feature/DatabaseUserTest.php` -> 15/15 green; three new tests cover the multi-host case, idempotent re-sync, and the Postgres no-duplicate regression. `CreateDatabaseUser`, `UpdateDatabaseUser`, and `DeleteDatabaseUser` already scope by `(username, host)`, so this just brings sync in line.

No migration. A composite unique index on `(server_id, username, host)` would be a natural follow-up but felt out of scope.

### DIFF
--- a/app/Actions/Database/SyncDatabaseUsers.php
+++ b/app/Actions/Database/SyncDatabaseUsers.php
@@ -28,10 +28,17 @@ class SyncDatabaseUsers
         foreach ($users as $user) {
             $databases = $user[2] != 'NULL' ? explode(',', $user[2]) : [];
 
+            $query = $server->databaseUsers()->where('username', $user[0]);
+
+            // MySQL/MariaDB distinguish users by (username, host); match both to avoid
+            // collapsing distinct accounts onto a single row. PostgreSQL's get-users-list
+            // returns an empty host (roles are host-agnostic), so only filter when present.
+            if ($user[1] !== '') {
+                $query->where('host', $user[1]);
+            }
+
             /** @var ?DatabaseUser $databaseUser */
-            $databaseUser = $server->databaseUsers()
-                ->where('username', $user[0])
-                ->first();
+            $databaseUser = $query->first();
 
             if ($databaseUser === null) {
                 $server->databaseUsers()->create([
--- a/tests/Feature/DatabaseUserTest.php
+++ b/tests/Feature/DatabaseUserTest.php
@@ -4,9 +4,12 @@ namespace Tests\Feature;
 
 use App\Enums\DatabaseUserPermission;
 use App\Enums\DatabaseUserStatus;
+use App\Enums\ServiceStatus;
 use App\Facades\SSH;
 use App\Models\Database;
 use App\Models\DatabaseUser;
+use App\Services\Database\Mysql;
+use App\Services\Database\Postgresql;
 use Illuminate\Foundation\Testing\RefreshDatabase;
 use Inertia\Testing\AssertableInertia;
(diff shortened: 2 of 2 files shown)


saeedvaziry:
@erhanurgun I appreceate the efforts on the issues and PRs however I rather to skip the long reads with AI generated too much details.

Lets keep the issues and PR descriptions for us humans with our own words.

Thanks

erhanurgun:
Sorry for the wall of text, you're right. Trimmed both the PR description and the issue.

PR #1079, opened 2026-05-03
[FIX] v3.x: Preserve import flash for non-database session drivers

### LINKED ISSUE #1080: [Bug] v3.x: Settings import shows no toast on non-database session drivers
### Installation Type

Local

### Vito Version

3.21.3

### Describe the Bug

Settings Import never shows a success or error toast on the default `redis` session driver, so I can't tell from the UI whether the import worked. I had to query the database to confirm it had actually run.

`VitoSettingController::import()` calls `config(['session.driver' => 'file'])` unconditionally. On redis, the flash payload ends up in the file store, the redirect reads from redis on the next request, finds nothing, and sonner renders nothing. The swap is fine when sessions live in the sqlite the import is overwriting (the `database` driver). For `redis`, `cookie`, `file`, `array` it just hides the result.

### Steps to Reproduce

1. Fresh install, default `SESSION_DRIVER` (redis).
2. `/admin/vito` -> Export, save the zip.
3. `/admin/vito` -> Import the same zip.
4. Page redirects, no toast.

### Expected Behavior

A success (or error) toast should render after import, like every other admin action.

### Logs

No errors in `storage/logs/laravel.log` or the browser console. The bug is the absence of feedback on a path that otherwise completes normally.

### Additional Context

Happy to send a PR. The fix is one conditional.

erhanurgun:
Closes #1080.

Default `SESSION_DRIVER` is `redis` (`config/session.php:21`), so any install on the defaults runs into this. The unconditional swap to the `file` driver writes the success flash to a store the next request never reads, so the redirect comes back without a toast and the user has no way to tell whether the import worked.

Limiting the swap to the `database` driver leaves every other driver untouched. Behavior on `database` is unchanged.

```
vendor/bin/phpunit tests/Feature/VitoSettingsTest.php
OK (3 tests, 11 assertions)
```

### DIFF
--- a/app/Http/Controllers/Admin/VitoSettingController.php
+++ b/app/Http/Controllers/Admin/VitoSettingController.php
@@ -91,7 +91,9 @@ class VitoSettingController extends Controller
         }
 
         // set session driver to file
-        config(['session.driver' => 'file']);
+        if (config('session.driver') === 'database') {
+            config(['session.driver' => 'file']);
+        }
 
         $request->validate([
             'backup_file' => 'required|file|mimes:zip',
--- a/tests/Feature/VitoSettingsTest.php
+++ b/tests/Feature/VitoSettingsTest.php
@@ -3,7 +3,11 @@
 namespace Tests\Feature;
 
 use Illuminate\Foundation\Testing\RefreshDatabase;
+use Illuminate\Http\UploadedFile;
+use Illuminate\Support\Facades\Artisan;
+use Illuminate\Support\Facades\File;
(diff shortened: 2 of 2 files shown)


erhanurgun:
Closing this. After deeper testing the fix does not address the observable behaviour: the success flash never reaches the next request even on the default `redis` driver and even with the swap untouched. The toast was missing for a different reason than the one I described, and I do not want to leave a misleading patch open. I will reopen if I can isolate the real cause.

PR #1110, opened 2026-05-22
[Feat] Server Provider Deletion

### LINKED ISSUE #1106: Show user confirmation/warning of deleting a server from server provider as well
### Discussed in https://github.com/vitodeploy/vito/discussions/1105

<sup>Originally posted by **aneeskhan47** May 19, 2026</sup>
Hey,

when i deleted the old server from vito (it also deleted it from digitalocean too because vito provisioned it in the first place so it makes sense) here my advice is can we add a confirm dialog for user to tell them would they like vito to delete api provider provisioned server from the provider itself or just from vito. This will make the user experience nice and user wouldn't accidentally delete servers from provider itself. This issue also exists in Ploi and Forge (idk if they still do this or ask user whether to delete from provider too).

This is really useful as sometimes a user only wants the server to be deleted from vito only and kept in the server provider itself.</div>

RichardAnderson:
This pull request introduces a major improvement to the server deletion workflow, allowing users to choose whether deleting a server should also destroy the underlying VM at the provider or only remove it from the application. This change is implemented across the backend, frontend, API documentation, and test suite to ensure a consistent and robust experience.

CleanShot 2026-05-22 at 10 24 02

Not shown for custom servers

Closes #1106

### DIFF
--- a/app/Actions/Server/DeleteServer.php
+++ b/app/Actions/Server/DeleteServer.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace App\Actions\Server;
+
+use App\Models\Server;
+use App\ServerProviders\Custom;
+use Illuminate\Support\Facades\Validator;
+use Illuminate\Validation\Rule;
+use Illuminate\Validation\ValidationException;
+
+class DeleteServer
+{
+    /**
+     * @param  array<string, mixed>  $input
+     *
+     * @throws ValidationException
+     */
+    public function delete(Server $server, array $input): void
+    {
+        $this->validate($server, $input);
+
+        if (array_key_exists('delete_from_provider', $input)) {
+            $server->deleteFromProvider = filter_var($input['delete_from_provider'], FILTER_VALIDATE_BOOLEAN);
+        }
+
+        $server->delete();
+    }
+
+    /**
+     * @param  array<string, mixed>  $input
+     */
+    private function validate(Server $server, array $input): void
+    {
+        $rules = [
+            'name' => [
+                'required',
+                Rule::in([$server->name]),
+            ],
+            'delete_from_provider' => [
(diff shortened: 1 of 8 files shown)


saeedvaziry:
Seems like the DeleteServer action never been used and validation is happening in the controller instead of the action

RichardAnderson:
@saeedvaziry - it does, but only for the UI interactions.  I tried to ensure backwards compatibility for the API, which is why it has validation in the controller for the API, where it defaults the "delete from provider" flag to true to replicate current functionality when not provided in the payload. 

Happy to make the API controller use the action as well - but this would be a breaking change for the API, requiring a new parameter to delete a server.

saeedvaziry:
> Happy to make the API controller use the action as well - but this would be a breaking change for the API, requiring a new parameter to delete a server.

it can be optional param for the API with default to true no?

RichardAnderson:
> > Happy to make the API controller use the action as well - but this would be a breaking change for the API, requiring a new parameter to delete a server.
> 
> it can be optional param for the API with default to true no?

This is what we have now.

**API delete**
delete_from_provider is optional, and if excluded, the default value is true
```php
    public function delete(Project $project, Server $server, Request $request): Response
    {
        $this->authorize('delete', [$server, $project]);

        $this->validateRoute($project, $server);

        Validator::make($request->all(), [
            'delete_from_provider' => ['nullable', 'boolean'],
        ])->validate();

        $server->deleteFromProvider = $request->boolean('delete_from_provider', true);
        $server->delete();
```

**UI controller Delete**
Requires both for non-custom servers
```php
      app(DeleteServer::class)->delete($server, $request->all());
```

saeedvaziry (approved):

PR #1112, opened 2026-05-22
[Feat] Github App Deployment

RichardAnderson:
Implements site deployments for GitHub App via short-lived installation tokens injected through a git credential helper, plus a handful of pre-existing-bug fixes uncovered along the way.

Implemented a git credential helper - clone and pull operations now run through a credential.helper for GitHub App deployments, token is per-repo and the token only ever lives in env at exec-time, never on disk.  Token is only valid for 1h.  Using a git credential helper here allows existing deployment scripts authenticate for any required commands during the deployment without changes. 

- Removed OS::runScript's info($command) call — it was writing the full deploy script (and now the token env var) to laravel.log at INFO
- AbstractSiteType::isolate() no longer guards createIsolatedUser on !userSharedWithSiblings() or !userExists() — the blade is fully idempotent and we need it to always run so existing isolated users (created before the auth-keys fix) get the server's public key installed into their authorized_keys.
- Removed userExists() and check-user-exists.blade.php — became zero-callers after the isolate() guard was dropped.
- SourceControl::siteValidationRules() now requires a Server — added user/project scoping to close a cross-tenant binding gap that became reachable once GitHub App was marked usableForSites(true).
- create-isolated-user.blade.php now writes the server's public key to the isolated user's authorized_keys — fixes a pre-existing bug where the terminal/live-console couldn't SSH into any isolated user.

### DIFF
--- a/app/Actions/GithubApp/HandlePushWebhook.php
+++ b/app/Actions/GithubApp/HandlePushWebhook.php
@@ -1,0 +1,52 @@
+<?php
+
+namespace App\Actions\GithubApp;
+
+use App\Jobs\Site\TriggerDeployFromWebhookJob;
+use App\Models\Site;
+use App\Models\SourceControl;
+
+class HandlePushWebhook
+{
+    /**
+     * @param  array<string, mixed>  $payload
+     */
+    public function handle(array $payload): void
+    {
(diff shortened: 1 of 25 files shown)


saeedvaziry:
Approved

PR #1121, opened 2026-05-24
[Feat] Hosted Domain Improvements

RichardAnderson:
Resolves the issue with validation of the domain over proxy by introducing a new short-lived HTTP/HTTPS validation method through /.well-known/vito.   Various other UI fixes / improvements around vhost generation. 

CheckDomainJob asks VerifyHostedDomain to confirm a domain. The verifier writes a short HMAC challenge file (keyed by APP_KEY + site/server/domain/random ID) to /var/lib/vito/verify/{key}/{id} on the server, then HTTP-fetches http://{domain}/.well-known/vito/{key}/{id} — pinning each DNS-resolved IP via CURLOPT_RESOLVE — and checks the response body matches. The vhost serves that path from disk via a location block rendered whenever the site has a verification_key. If any IP returns the right body, the domain activates; otherwise it goes pending. Another Vito install can't fake it because they don't have your APP_KEY.


saeedvaziry:
Approved

PR #1177, opened 2026-06-19
Fix env secrets being lost after pulling from server and saving

saeedvaziry:
Make the live server .env file the source of truth when restoring masked secrets on save and pull, so
  secrets that drifted out of band are no longer silently overwritten with a stale DB copy.

## Release Notes

* **Refactor**
  * Improved environment variable processing to preserve secret values from the live server when secret fields are submitted empty, and to keep secret status consistent.
  * Updated worker environment updates to follow the same secret-preservation rules.

* **Tests**
  * Added/updated integration tests to verify only secret keys are stored, secret values are masked on read, and out-of-band secret rotations are respected during updates.
  * Expanded unit tests to cover live-based secret merge behaviour, including preservation vs non-preservation for empty values.

### DIFF
--- a/app/Actions/Site/GetEnv.php
+++ b/app/Actions/Site/GetEnv.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace App\Actions\Site;
+
+use App\Helpers\EnvParser;
+use App\Models\Site;
+
+class GetEnv
+{
+    /**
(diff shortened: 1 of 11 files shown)


saeedvaziry:
@coderabbitai review

RichardAnderson (approved):

PR #1187, opened 2026-06-27
Import Feature of WorkFlow & Delete Alias

fatihalp:
Workflow is a fantastic and highly powerful feature. However, using and managing them can be difficult.
WorkFlow Import is only works (best & tested) with Hetzner.

I should be able to create Wordflows on a project basis using files; I mentioned this feature before.

Additionally, the domain addition feature isn't working properly, which is why I removed it from the project.
If you need create new site, it's cloud.


RichardAnderson:
Hi @fatihalp,

Thanks for the PR, and for contributing to Vito!   It seems like there are a lot of changes in this PR unrelated to your description, including the removal of site aliases, etc.   It's also targeting 3.x, however, we are tracking for a 4.x release next week, and there are substantial conflicts between these updates and the updates in 4.x

Would it be possible to break down these features into smaller, self-contained PRs against the 4.x codebase?

For note;
- The domain alias functionality has been replaced in v4 for Hosted Domains.  This improves its usability, and simplifies the generation of SSL certificates and associated vhost blocks.
- Workers have been overhauled, including the ability to sync worker states back from the server.  While this does not bring in workers maintained outside of vito, it does provide better support for workers managed within vito.
- Site creation, server creation and service installation has been streamlined - this includes an updated suite of supported database versions. 
- The UI for Server Providers in the creation of a new server has been overhauled as well, more information is being pulled back to show live availability of plans and associated costs. 
- The timeout for SSL connections for slower server builds has been increased.
- .... many more improvements.

fatihalp:
Hi @RichardAnderson 

I initially built these features to fulfill my own needs. I wanted to open a pull request to share those needs with the community. I already figured you'd want to do this in v4 anyway.

Thanks for checking it out. I'd love to implement these features once v4 becomes stable. Looking forward to the new version! :)

I actually built a similar panel called 'purepanel' back in the day, but I couldn't manage to maintain it over time. I really want to support this project both technically and, as soon as my financial resources allow, financially as well.

Issue #933, opened 2025-11-15
Mysql as Vito's default database


saeedvaziry:
Vito uses SQLite for its database driver.

Since version 3, Vito received a lot of features and the number of write attempts to the database has increased significantly.

This ends up causing "database is locked" errors when multiple write attempts are made simultaneously.

Migrating to Mysql will solve this issue.

heyjorgedev:
Hi @saeedvaziry, have you tried to enable WAL mode on sqlite before doing such migration?

there are params like that could reduce or even eliminate the locks entirely.

```sql
PRAGMA journal_mode = WAL;
PRAGMA busy_timeout = 5000;
etc...
```

Don't know if you can reproduce the database is locked issue somehow but I will submit a PR with simple changes that should improve the locks.

btw, vito is amazing

saeedvaziry:
Its a good idea but WAL creates additional files and during time can slow down the read queries and might be challenging to maintain in a self-hosted app

heyjorgedev:
Hi @saeedvaziry,

I can see your points, from my experience I can tell you that reads are actually faster using the WAL, journal_mode WAL creates 2 more files, the WAL and a shared memory file to help concurrency. so 3 files in total (and not at all times, when a checkpoint happens and the WAL becomes empty sqlite actually deletes the WAL).

The WAL is like an append-only log to keep reads and writes separated helping the sqlite file to not be locked by writers.

You're the boss, I honestly think that it is worth a try before including another daemon on the server that needs to be taken care of, spends memory on the server and a backup is not just copying a file.

saeedvaziry:
@heyjorgedev you're welcome for a PR of course. Vito already supports mysql but it is still my first prio to keep the instance itself as light as possible to be able to boot up vito on low resources

wawans:
I think adding `php8.4-mysql php8.4-pgsql`  in `Dockerfile` install script should be enough, then we can add environment variables `DB_CONNECTION=mysql`  or pgsql.

PR #1189, opened 2026-07-01
[Feat] Support MySQL and MariaDB as the panel database (#933)

RauchenwaldC:
SQLite hits "database is locked" under the panel's concurrent background writes (#933). This lets Vito's own data run on MySQL or MariaDB while keeping SQLite the zero-config default. The full test suite is green on SQLite, MySQL 8.4 and MariaDB 11 (serial and --parallel).

- config/database.php: add a `mariadb` connection. MariaDB needs Laravel's dedicated `mariadb` driver -- the `mysql` driver compiles JSON-column equality to `cast(? as json)`, which MariaDB rejects.
- GetMetrics: make the monitoring query portable -- driver-aware date bucketing (strftime vs DATE_FORMAT), backtick-quote the reserved word `load`, and aggregate `created_at` (MIN) for ONLY_FULL_GROUP_BY.
- PluginTest: use delete() instead of truncate() -- TRUNCATE on an FK-referenced table errors 1701 on MySQL/MariaDB, and its implicit commit breaks RefreshDatabase isolation.
- Fix two engine-exposed test fragilities: JSON object-key ordering (assertEquals, not assertSame) and a cross-table auto-increment id collision in assertJsonMissing (scope the check to the response data ids).
- scripts/install.sh: prompt for the panel DB engine and provision MySQL/MariaDB (writing the correct DB_CONNECTION for each).
- Add a `migrate-from-sqlite-to-mysql` command to move an existing SQLite install to MySQL or MariaDB.
- CI: add a sqlite/mysql/mariadb test matrix with service containers.
- .env.example + install/configuration docs: document the option.

## Summary by CodeRabbit

* **New Features**
  * Added MariaDB as a supported database engine, including installer and environment configuration support.
  * Added a command to migrate data from SQLite to MySQL/MariaDB.
* **Documentation**
  * Updated installation and configuration guides with database engine selection, environment variables, and migration workflow.
  * Expanded the environment template with database switching instructions.
* **Bug Fixes**
  * Improved cross-database metrics time bucketing and formatting consistency.
* **Tests & CI**
  * Expanded CI to run the test suite for SQLite, MySQL, and MariaDB, and refined related assertions.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -8,6 +8,16 @@ APP_URL=
 #WS_BROADCAST_SECRET=
 WS_ALLOWED_ORIGINS="${APP_URL}"
 
+# Database — SQLite is the default and needs no configuration.
+# To use MySQL or MariaDB, set DB_CONNECTION and uncomment the block below.
+# Use "mariadb" (not "mysql") for MariaDB so Laravel emits MariaDB-compatible SQL.
+#DB_CONNECTION=mysql   # or "mariadb" for MariaDB
+#DB_HOST=127.0.0.1
+#DB_PORT=3306
+#DB_DATABASE=vito
+#DB_USERNAME=vito
+#DB_PASSWORD=
+
 FILESYSTEM_DRIVER=local
 
 MAIL_MAILER=smtp
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,6 +7,7 @@ on:
       - 1.x
       - 2.x
       - 3.x
+      - 4.x
   pull_request:
 
 jobs:
@@ -14,10 +15,40 @@ jobs:
     runs-on: ubuntu-22.04
 
     strategy:
-      fail-fast: true
+      fail-fast: false
       matrix:
         php: [ 8.4 ]
         node-version: [ "22.x" ]
+        database: [ sqlite, mysql, mariadb ]
+
+    services:
+      mysql:
+        image: mysql:8.4
+        env:
+          MYSQL_ALLOW_EMPTY_PASSWORD: "true"
+          MYSQL_DATABASE: vito_test
+        ports:
+          - 3306:3306
+        options: >-
+          --health-cmd="mysqladmin ping --silent"
+          --health-interval=10s
+          --health-timeout=5s
+          --health-retries=10
+
+      mariadb:
+        image: mariadb:11
+        env:
+          MARIADB_ALLOW_EMPTY_ROOT_PASSWORD: "true"
+          MARIADB_DATABASE: vito_test
+        ports:
+          - 3307:3306
+        options: >-
+          --health-cmd="healthcheck.sh --connect --innodb_initialized"
+          --health-interval=10s
+          --health-timeout=5s
+          --health-retries=10
+
+    name: tests (${{ matrix.database }})
 
     steps:
       - uses: actions/checkout@v4
@@ -41,6 +72,7 @@ jobs:
         run: composer install --prefer-dist --no-progress --no-suggest
 
       - name: Create sqlite database
+        if: matrix.database == 'sqlite'
         run: touch storage/database-test.sqlite
 
       - name: Set up the .env file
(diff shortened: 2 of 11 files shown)


RauchenwaldC:
Addressed 7 of the 8:

- **`migrate:fresh` exit code (Critical):** now captured; the command aborts before copying any data if the schema build fails.
- **Config-cache cut-over (Major):** the command refuses to run when config is cached (`App::configurationIsCached()`) and tells you to `php artisan config:clear` first — cached config means `.env` isn't loaded, so both the target-connection read and the `DB_CONNECTION` rewrite would otherwise silently no-op. Documented the prerequisite too.
- **`GetMetrics` bucket label (Major):** each group is now labelled by the `date_interval` bucket boundary instead of a sample timestamp.
- **`V_DB_ENGINE` validation (Major):** installer rejects unsupported values up front.
- **`V_DB_PASSWORD` safety (Major):** default is now `openssl rand -hex 16`, and an operator-supplied value is validated against `^[A-Za-z0-9._-]+$` before it's embedded in the SQL heredoc / `.env`.
- **`.env.example` (Minor):** the example now shows the `mariadb` option.
- **Docs "every table" (Minor):** clarified that transient framework tables (`migrations`, `sessions`, `cache`, `jobs`, `failed_jobs`) are intentionally skipped.

**Not applied — extract into an Action (Major, maintainability):** I kept it as a self-contained command to mirror its direct counterpart `MigrateFromMysqlToSqlite`, which is likewise an inline console command, and because the logic is tightly coupled to interactive console I/O (`components->info/warn/twoColumnDetail`, `confirm`). Happy to extract into `app/Actions/` if you'd prefer consistency with the guideline over consistency with the existing sibling.

Suite still green on SQLite, MySQL 8.4 and MariaDB 11 (1762 passed); Pint + PHPStan clean.

RauchenwaldC:
4 applied / 3 respectfully skipped:

**Applied**
- **Stream the SQLite read:** `copyTable()` now streams via a cursor and inserts in 500-row batches, so a large table (e.g. `metrics`) is never fully materialised.
- **Abort on missing target table:** an unexpected source table now aborts by default; added `--skip-missing` to opt into skipping.
- **Quiesce before copy:** the command warns to stop the web server / queue workers / scheduler before copying, and the docs state it as a required step. I didn't auto-toggle maintenance mode — it doesn't stop queue workers, and driving `down`/`up` around a DB cut-over is fragile and environment-specific, so an explicit precondition is safer.
- **Installer fail-fast:** the MySQL/MariaDB `apt install`, `systemctl enable --now`, and provisioning heredoc now abort on failure (matching the script's existing `if ! …; then exit 1` style), so `.env` is only written after a successful DB setup.

**Skipped (with reasons)**
- **Digest-pin the service images:** nothing in this repo is digest-pinned (workflows use `@v4`/`@v2` action tags and tag-based images); pinning one workflow's service images would be inconsistent and add manual-bump maintenance. `mysql:8.4` / `mariadb:11` are standard for CI service containers.
- **Normalize matrix inline-array spacing:** there's no yamllint config/step in CI, and `[ sqlite, mysql, mariadb ]` matches the spacing of the existing `php:` / `node-version:` matrices in the same file.
- **Fail-fast in `GetMetrics` for pgsql/sqlsrv:** the panel DB is sqlite/mysql/mariadb only (it was sqlite-only before this PR); `pgsql`/`sqlsrv` are default-scaffold connections Vito never uses for its own DB, and nothing else in the app guards against them — a metrics-only guard would be inconsistent and out of scope.

Suite still green on all three engines; Pint + PHPStan clean.

saeedvaziry:
Please for such refactors first start a discussion in Discord or in the Github Discussions.

Vito remains on SQLite, however you can migrate your own instance to use mysql

RauchenwaldC:
@saeedvaziry closing this PR directly contradicts your statement in the issue #933 that this PR attempt to resolve.

See https://github.com/vitodeploy/vito/issues/933#issuecomment-3567870670
> you're welcome for a PR of course. 

In that case, nevermind.

saeedvaziry:
I am aware. I tried to migrate for v4 but some issues stopped me continuing. I might again have an attempt for it for v5 maybe.

But for now I will rather try to support mysql as alternative option not the default one.

So feel free to open a pr if something doesn't work with mysql in the current setup.

Thanks

PR #1207, opened 2026-07-23
[Fix] Prevent backups:run crash for backups whose server was deleted

### LINKED ISSUE #1206: [Bug]: backups:run crashes on every run after deleting a server that had backups
### Installation Type

Self-hosted (VPS)

### Vito Version

4.0.1

### Describe the bug

If I delete a server that has an enabled database backup, its backup rows are left behind with a `server_id` that no longer exists. `Server::deleting` (`app/Models/Server.php`) cleans up sites, services, databases and so on, but not `backups()`, so they're orphaned.

From then on the scheduled `backups:run` fails on every run:

```
Attempt to read property "project_id" on null
Scheduled command ['php' 'artisan' backups:run] failed with exit code [1].
```

`RunBackup::run()` reads `$backup->server->project_id` (`app/Actions/Backup/RunBackup.php:33`), and the retention hook does the same via `ManageBackupFile::delete()` (`app/Actions/Backup/ManageBackupFile.php:37`) — both on a server that is now `null`.

### Steps to reproduce

1. Add a server and create an enabled database backup on it (e.g. hourly).
2. Delete the server.
3. Wait for the next scheduled run, or run `php artisan backups:run`.
4. It exits 1 and logs the null-property error, every time.

### Expected Behavior

Deleting a server should also remove its backups (as it already does for databases), and `backups:run` should skip a backup whose server is gone instead of crashing.

### Logs

```
production.ERROR: Attempt to read property "project_id" on null
production.ERROR: Scheduled command ['/usr/bin/php8.4' 'artisan' backups:run] failed with exit code [1].
```

### Additional context

Still reproduces on 4.x after #1178: `backups:reconcile` recovers stuck jobs, but a server deleted while it still has live backups orphans them and crashes the runner on every tick.

erhanurgun:
Closes #1206

When a server that has backups is deleted, its `Backup` rows are left with a dangling `server_id` (there's no FK on the column). The scheduled `backups:run` then reads `$backup->server->project_id` on a null server and crashes on every run.

This makes the runner and the backup-file cleanup resilient to those already-orphaned rows:
- `RunBackupCommand`: skip backups whose server is gone (`whereHas('server')`) and eager-load `server` (also fixes a pre-existing N+1).
- `ManageBackupFile::delete()`: null-safe when the server is missing — it logs and drops the local record instead of dispatching an SSH job, and skips the socket broadcast. (Server loaded via `Server::find()` so PHPStan doesn't treat the non-nullable relation as never-null.)

Also removed the unused `ManageBackupFile::download()` method (dead since the Inertia migration in #593 — its only caller was the old Filament widget, which wasn't re-wired), per review.

Note: cascading a server's backups on delete is now handled by #1198, so this PR doesn't touch that — it only covers rows already orphaned on existing installs, which #1198's cascade doesn't reach.

Tests (SSH::fake + RefreshDatabase):
- `RunBackupCommandTest::test_does_not_run_backups_whose_server_is_missing`
- `BackupTest::test_delete_orphaned_backup_file_hard_deletes_without_dispatching_job`

No migration.

## Summary by CodeRabbit

* **Bug Fixes**
  * Backup file deletions are now safely handled when the associated server is missing, hard-deleting the file without triggering remote deletion.
  * Backups scheduled with a missing server are skipped, and no jobs/events are dispatched.
* **Tests**
  * Added feature coverage for hard-deleting an orphaned backup file without dispatching a delete job.
  * Added unit coverage ensuring the `backups:run` command does not start backups for missing servers.


erhanurgun:
Heads up: I noticed #1198 already adds the `Server::deleting` backup cascade (and reworks `backups()` into a `HasManyThrough`), so I trimmed that hunk out of this PR to avoid a conflicting duplicate. This one now only makes `backups:run` and the backup-file ops resilient to backups that are *already* orphaned on existing installs, which #1198's future-only cascade doesn't cover (I confirmed it touches neither `RunBackupCommand` nor `ManageBackupFile`). Happy to reorder behind #1198, fold the runner guard in, or re-add the cascade here if #1198 stalls, whatever you prefer.

RichardAnderson:
Hi @erhanurgun - I just merged the approved PR #1198 - and these changes complement it well, this has introduced a conflict on one of the tests.

erhanurgun:
Thanks @RichardAnderson - glad it fits. I've merged #1198 in and resolved the `BackupTest` conflict (kept both the orphan-path tests and your global-backup tests). Green on my side again.

app/Actions/Backup/ManageBackupFile.php
```diff
@@ -19,7 +22,12 @@ class ManageBackupFile
      */
     public function download(BackupFile $file): StreamedResponse
```
RichardAnderson: It does not appear that anything (other than the new test you added) actually uses download anymore.  I believe we had some download backups functionality in v2 of Vito, but it does not actually appear it was ported to v3/v4, so this is actually dead code now.

We should likely delete this (and the new test) instead of updating something that is never used.

tests/Feature/BackupTest.php
```diff
@@ -603,6 +606,45 @@ public function test_cannot_enable_a_backup_being_deleted(): void
...
+        ]);
+        $file = BackupFile::factory()->create([
+            'backup_id' => $backup->id,
+            'status' => BackupFileStatus::CREATED,
+        ]);
+
+        $this->expectException(RuntimeException::class);
+
+        app(ManageBackupFile::class)->download($file);
+    }
```
RichardAnderson: See comment above, I think we can remove this test and the entire download function.

app/Actions/Backup/ManageBackupFile.php
erhanurgun: Good catch - verified it: `download()` has been dead since the Inertia migration (#593). Its only caller was the old Filament `BackupFilesList` widget, which was removed then and never re-wired (no route/controller/frontend/API references it). Removed `download()` and its test, so this PR is now just the `delete()` null-safety plus the `backups:run` server guard. Thanks for the review!

tests/Feature/BackupTest.php
erhanurgun: Done - removed this test along with the `download()` method.

app/Models/Server.php
```diff
@@ -144,6 +144,11 @@ public static function boot(): void
                     $log->delete();
                 });
                 $server->services()->delete();
+                $server->backups()->each(function ($backup): void {
+                    /** @var Backup $backup */
+                    $backup->files()->delete();
```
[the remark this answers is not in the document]
erhanurgun: This targeted the `Server::deleting` cascade hunk I dropped from this PR - #1198 now owns that cascade, so it's out of scope here. One thing I did notice while checking though: the cascade deletes the DB rows, but `BackupFile` has no `deleting`/`deleted` hook, so `deleteFile()` never runs during a server delete and the remote objects (S3/Dropbox/FTP) look like they'd be left behind. Not this PR's code to fix, but happy to open a separate issue if useful.

tests/Unit/Commands/RunBackupCommandTest.php
```diff
@@ -121,4 +122,19 @@ public function test_runs_enabled_backup_even_after_a_failed_run(): void
...
+        Bus::fake();
+        Carbon::setTestNow('2026-06-19 10:00:00');
+
+        $backup = $this->createBackup(['interval' => '0 * * * *', 'server_id' => 999999]);
+
+        $this->artisan('backups:run')
+            ->expectsOutput('0 backups started');
+
+        $this->assertDatabaseHas('backups', ['id' => $backup->id]);
+        Bus::assertNothingDispatched();
```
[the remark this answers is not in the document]
erhanurgun: Done - added `'server_id' => 999999` to the assertion.

RichardAnderson (approved):

PR #1208, opened 2026-07-23
[Fix] Wait for cloud-init and apt

### LINKED ISSUE #1205: [Bug]: Installation fails at 15% every time
### Installation Type *

VPS

### Vito Version *

4.0.1

### Describe the Bug *

I have created a DO server provider and provided a token with FULL ACCESS.

When I try and create a DO server, it creates the droplet but fails at 15% every time.

### Steps to Reproduce *

Try creating a DO server via VitoDeploy.

### Expected Behavior *

Droplet should be created and VitoDeploy should install successfully.

### Logs *

E: 
Could not get lock /var/cache/apt/archives/lock. It is held by process 1408 (apt-get)
E: Unable to lock directory /var/cache/apt/archives/

SSH command failed with an error

### Screenshots

_No response_

### Additional Context

_No response_

RichardAnderson:
Wait for cloud-init and apt locks to clear before provisioning starts, so new servers (notably DigitalOcean) no longer fail with "Could not get lock /var/cache/apt/archives/lock".

Closes #1205

New command executed at the start:
image

Server continues past the upgrade process:
image

DO 1CPU/1GB RAM VPS provisions successfully (where it previously failed):
image

## Summary by CodeRabbit

* **New Features**
  * Server installation now advances to **“preparing-system”** and then waits for the operating system to complete initial boot readiness.
  * Provisioning now pauses until first-boot prerequisites finish, including:
    * waiting for `cloud-init` when it’s in the running state, and
    * waiting for package manager lock releases (via `fuser`).
  * Added timeout-based safeguards for both boot readiness and provisioning checks, with clear error messaging when limits are exceeded.

### DIFF
--- a/app/Actions/Server/InstallServer.php
+++ b/app/Actions/Server/InstallServer.php
@@ -54,6 +54,8 @@ class InstallServer
      */
     public function install(): void
     {
+        $this->progress(5, 'preparing-system');
+        $this->server->os()->waitForBoot();
         $this->createUser();
         $this->progress(15, 'installing-updates');
         $this->server->os()->upgrade();
--- a/app/SSH/OS/OS.php
+++ b/app/SSH/OS/OS.php
@@ -26,6 +26,19 @@ class OS
     /**
      * @throws SSHError
      */
+    public function waitForBoot(int $timeout = 300): void
+    {
+        $this->server->ssh()->exec(
+            view('ssh.os.wait-for-boot', [
+                'timeout' => $timeout,
+            ]),
+            'wait-for-boot'
+        );
+    }
+
+    /**
+     * @throws SSHError
+     */
     public function installDependencies(): void
     {
         $this->server->ssh()->exec(
--- a/resources/views/ssh/os/wait-for-boot.blade.php
+++ b/resources/views/ssh/os/wait-for-boot.blade.php
@@ -1,11 +1,14 @@
 echo "Waiting for cloud-init and apt to finish first-boot tasks..."
 
 if command -v cloud-init >/dev/null 2>&1; then
-    sudo timeout {{ $timeout }} cloud-init status --wait >/dev/null 2>&1
-    __vito_ci_exit=$?
-    if [ "$__vito_ci_exit" -eq 124 ]; then
-        echo "Timed out after {{ $timeout }}s waiting for cloud-init to finish." >&2
-        exit 1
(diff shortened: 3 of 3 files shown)


saeedvaziry:
@RichardAnderson do we need to apply this to all servers

RichardAnderson:
> @RichardAnderson do we need to apply this to all servers

At the moment its only effecting DO servers (confirmed even the larger servers have this, but it's a worthwhile check regardless, and could solve problems should other providers introduce default cloud-init scripts that run apt updates/upgrades automatically.

saeedvaziry:
Thanks for tackling the provisioning race here. I spotted two edge cases worth fixing before merge:

- The `cloud-init status --wait` command captures every non-zero exit code, but only handles `124` (the normal timeout case). If cloud-init itself fails, or `timeout` cannot run the command (`125`–`127`), the script carries on and eventually says the system is ready. After keeping the clearer message for `124`, it should stop on any other non-zero result too.
- If `fuser` is not installed, the script only prints a warning and then continues with “System is ready for provisioning.” That means the original apt-lock failure can still happen on an image without `fuser`. It would be safer to use a fallback lock check or fail at that point instead of proceeding without checking.

The changed PHP lines have 100% diff coverage in CI; these are both about making the shell script fail safely when it cannot guarantee readiness.

RichardAnderson:
> Thanks for tackling the provisioning race here. I spotted two edge cases worth fixing before merge:
> 
> * The `cloud-init status --wait` command captures every non-zero exit code, but only handles `124` (the normal timeout case). If cloud-init itself fails, or `timeout` cannot run the command (`125`–`127`), the script carries on and eventually says the system is ready. After keeping the clearer message for `124`, it should stop on any other non-zero result too.
> * If `fuser` is not installed, the script only prints a warning and then continues with “System is ready for provisioning.” That means the original apt-lock failure can still happen on an image without `fuser`. It would be safer to use a fallback lock check or fail at that point instead of proceeding without checking.
> 
> The changed PHP lines have 100% diff coverage in CI; these are both about making the shell script fail safely when it cannot guarantee readiness.

Hi @saeedvaziry - this is the expected behaviour.  The purpose of implementation is as a light safety gate to ensure that if we can detect these things we do, but if we can't or get an unexpected error, we don't block the provisioning of the server.   Given this, this check is not a hard-gate on the process, but rather a soft-check, picking up a common issue with some cloud providers that run apt-update via cloud-init or via a custom process post-provision, without blocking the whole process if cloud-init is not working, etc.

saeedvaziry (approved):

PR #1209, opened 2026-07-24
[Feat] Networks: private networking via WireGuard, custom, and provider-synced VPCs

RichardAnderson:
Adds project-scoped private networks that link servers either over Vito-managed WireGuard tunnels (with peer devices for laptops and CI runners), over private IPs you already have, or by discovering and syncing existing VPCs from AWS, DigitalOcean, Hetzner, Linode, and Vultr, each network carrying its own firewall rules that Vito keeps applied across every member.

## View networks
image

## View network details
image

## Summary by CodeRabbit

- **New Features**
  - Added full Networks management (WireGuard, Custom, and provider-discovered networks) including servers, peers, firewall rules, and automatic reconciliation/sync with status tracking and retries.
- **Documentation**
  - Added/expanded Networks docs for overview, creating networks, peers, servers, firewalls, and provider networks, plus automation notes.
- **Bug Fixes**
  - Improved firewall rule application to correctly handle empty protocol/port values, improved managed-network firewall visibility, and enhanced navigation “active” highlighting (including exact-match routes).


saeedvaziry:
• ## Review result: request changes

  ### Specs

  - P1 — IPv6 endpoints generate invalid WireGuard configs. app/Services/VPN/WireGuard.php:166 concatenates addresses as $ip:$port; IPv6 requires
    [address]:port. Downloaded peer configs have the same issue in app/Actions/Network/GetNetworkPeerConfig.php:56. Networks containing an IPv6-addressed
    server will fail synchronization.

  - P2 — Removing the last provider-backed server strands the network. app/Actions/Network/SyncProviderNetworks.php:347 discards provider contexts without
    current servers. After the sole managed server is deleted, sync never prunes the empty network, while its provider association prevents manual
    deletion.

  - P2 — Firewall protocol and port cannot be independently optional. app/Actions/Network/ManageNetworkFirewallRule.php:93 uses reciprocal required_with
    rules. This rejects “all protocols on port 3306” and “UDP on all ports,” although both the form and documentation describe protocol and port as
    independently optional.

  - P2 — Custom-network addresses need CIDR containment validation. app/Actions/Network/CreateNetwork.php:216 verifies ownership, type, and uniqueness,
    but not that the chosen address belongs to the declared CIDR. Add and update have the same omission. A mismatched address breaks connectivity while
    potentially opening the firewall to an unrelated range.

  - P3 — Provider name is absent from network settings. app/Http/Middleware/HandleInertiaRequests.php:94 does not load serverProvider, while
    NetworkResource only emits provider when that relation is loaded. Managed network settings therefore display -.

  ### Standards and security

  - P1 — Unvalidated server IP reaches an SSH shell template. app/Actions/Server/EditServer.php:49 resynchronizes peers before checking the new address,
    but validation only requires a string. The value becomes an unquoted UFW source through app/Actions/Network/MaterializeServerNetworkRules.php:209,
    permitting shell injection on ready sibling servers. Validate it as an IP address before saving or dispatching synchronization.

  - P1 — Server deletion leaves stale allow rules on remaining members. app/Models/Server.php:141 only reconciles WireGuard networks. For custom networks
    without a CIDR—and provider networks with per-member rules—the remaining servers retain /32 allow rules for the deleted server’s address. That address
    may later be reassigned. The reconciliation should cover every affected network type and live in an Action rather than model callbacks.

  - P2 — Provider failures are swallowed. app/Actions/Network/SyncProviderNetworks.php:48 logs provider failures and continues, so the job reports success
    and users receive no failure state. This conflicts with the project rule that provider errors must bubble. Multiple failures can be collected and
    rethrown after processing unaffected connections.

saeedvaziry (approved):

